PR #49, opened 2021-03-13
Change config and default

ota-meshi:
- Change `plugin:regexp/recommended` config
  - Add `regexp/confusing-quantifier` rule.
  - Add `regexp/control-character-escape` rule.
  - Add `regexp/negation` rule.
  - Add `regexp/no-dupe-disjunctions` rule.
  - Add `regexp/no-empty-alternative` rule.
  - Add `regexp/no-empty-capturing-group` rule.
  - Add `regexp/no-invalid-regexp` rule.
  - Add `regexp/no-lazy-ends` rule.
  - Add `regexp/no-legacy-features` rule.
  - Add `regexp/no-non-standard-flag` rule.
  - Add `regexp/no-obscure-range` rule.
  - Add `regexp/no-optional-assertion` rule.
  - Add `regexp/no-potentially-useless-backreference` rule.
  - ~~Add `regexp/no-standalone-backslash` rule.~~
  - Add `regexp/no-super-linear-backtracking` rule.
  - Add `regexp/no-trivially-nested-assertion` rule.
  - Add `regexp/no-trivially-nested-quantifier` rule.
  - Add `regexp/no-unused-capturing-group` rule.
  - Add `regexp/no-useless-assertions` rule.
  - Add `regexp/no-useless-character-class` rule.
  - Add `regexp/no-useless-dollar-replacements` rule.
  - Add `regexp/no-useless-escape` rule.
  - Add `regexp/no-useless-flag` rule.
  - Add `regexp/no-useless-lazy` rule.
  - Add `regexp/no-useless-non-capturing-group` rule.
  - Add `regexp/no-useless-quantifier` rule.
  - Add `regexp/no-useless-range` rule.
  - Add `regexp/no-zero-quantifier` rule.
  - Add `regexp/optimal-lookaround-quantifier` rule.
  - Add `regexp/optimal-quantifier-concatenation` rule.
  - Add `regexp/prefer-character-class` rule.
  - Add `regexp/prefer-predefined-assertion` rule.
  - Add `regexp/prefer-range` rule.
  - Add `regexp/prefer-unicode-codepoint-escapes` rule.
  - Add `regexp/sort-flags` rule.
  - Add `regexp/strict` rule.
  - Add `no-empty-character-class` rule.
  - Remove `no-invalid-regexp` rule.

- Change default option
  - Change `hexadecimalEscape` option of `regexp/letter-case` rule from `ignore` to `lowercase`.
  - Change `controlEscape` option of `regexp/letter-case` rule from `ignore` to `uppercase`.


RunDevelopment:
I think there are 2 problems with the recommended config:

1. `regexp/strict` overlaps with 3 other rules. `regexp/strict` is a superset of `regexp/no-standalone-backslash`, `regexp/no-octal`, and `regexp/no-useless-escape`.

    I don't know whether we should remove those 3 rules from the rec config or whether we should live with double reporting in those cases.

2. `regexp/no-useless-non-capturing-group`: I love this rule but it's so annoying when writing regexes. It always tries to remove non-capturing groups and overshadows other errors while I write my regexes. It's great for code review but not so much for live editing.

    Maybe we should remove it from the rec config? I don't want people to install our plugin only to be annoyed by this rule.

    Another idea would be to make the rule less annoying. The problem is that it underlines the whole group. Maybe we could make the report range smaller (e.g. only the `(?:` part)? That way the rule wouldn't prevent me from seeing errors within the group.

ota-meshi:
> regexp/strict overlaps with 3 other rules. regexp/strict is a superset of regexp/no-standalone-backslash, regexp/no-octal, and regexp/no-useless-escape.

I think it's good to turn off `regexp/no-standalone-backslash`. It rule is a complete subset of `regexp/strict`.

However, there seems to be a slight difference between `regexp/no-useless-escape` and `regexp/no-octal` from `regexp/strict`.

```js
var foo = /[\|]/; // Reported in `regexp/no-useless-escape`. But not reported in `regexp/strict`.
var foo = /[\1-\4]/; // Reported in `regexp/strict`. But not reported in `regexp/no-octal`.
```

These rules may report duplicates, but I think it's good to turn them on.

RunDevelopment:
> However, there seems to be a slight difference between `regexp/no-useless-escape` and `regexp/no-octal` from `regexp/strict`.
> 
> ```js
> var foo = /[\|]/; // Reported in `regexp/no-useless-escape`. But not reported in `regexp/strict`.
> var foo = /[\1-\4]/; // Reported in `regexp/strict`. But not reported in `regexp/no-octal`.
> ```
> 
> These rules may report duplicates, but I think it's good to turn them on.

I agree with you on `regexp/no-useless-escape` but I think we should remove `regexp/no-octal` from recommended. `regexp/strict` disallows all octal escapes while `regexp/no-octal` does allow some octal escapes. So `regexp/no-octal` is a subset of `regexp/strict`.

ota-meshi:
> I think we should remove regexp/no-octal from recommended.

You're right. I will change this PR.

RunDevelopment (approved):

PR #73, opened 2021-03-23
Fix false positives for `match` with `g` flag in `regexp/prefer-regexp-test` rule

ota-meshi:


fisker:
Hey, I didn't know there is a same rule here, when I fix the same bug in `eslint-plugin-unicorn`, I excluded the regex literals with `g` in my fix https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1173.

And have checked do we have [this same bug](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1150) here?

RunDevelopment:
> we have this same bug here?

Yes, we do. I'll open an issue.

PR #95, opened 2021-04-09
Add `regexp/no-empty-alternative` rule

ota-meshi:


RunDevelopment:
Is it really necessary to move this over?

We already have `regexp/prefer-question-quantifier`. I'd rather make this rule smarter instead of adding a new one. What do you think?

ota-meshi:
I think these two rules have different aims.
`no-empty-alternative` helps find mistakes or hints that may be mistakes.
`prefer-question-quantifier` helps to enforce a consistent style.
The user can use either or both.

Also, if we want to avoid duplicate checks, I think we may consider removing the "empty alternative" check from the `prefer-question-quantifier`.

How do you think the `prefer-question-quantifier` will change?

RunDevelopment:
> I think these two rules have different aims.
> `no-empty-alternative` helps find mistakes or hints that may be mistakes.
> `prefer-question-quantifier` helps to enforce a consistent style.

Good point. Then let's add `no-empty-alternative`.

> Also, if we want to avoid duplicate checks, I think we may consider removing the "empty alternative" check from the `prefer-question-quantifier`.

No, I think it's good to have a fixable "empty alternative" check in `prefer-question-quantifier`.

RunDevelopment (approved):

PR #96, opened 2021-04-09
Add `regexp/no-lazy-ends` rule

ota-meshi:


RunDevelopment (approved):

RunDevelopment (changes requested):
Actually, we should think about naming here.

I always used "lazy" to refer to non-greedy (= lazy) quantifiers. However, `eslint-plugin-regexp` has the `no-useless-non-greedy` rule. We should name rules consistently, so one of them has to be renamed.

Do we choose "lazy" or "non-greedy"?

I personally prefer "lazy" simply because it's shorter. However, I will also understand if you do not want to rename your existing rule.

ota-meshi:
I'm not familiar with English, so I often take words from MDN.
`non-greedy` referred to the following link:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Quantifiers#types

I think `lazy` is a good word.
If we need a consistent naming, I think we need to change `no-useless-non-greedy`.

RunDevelopment:
> I think `lazy` is a good word.
> If we need a consistent naming, I think we need to change `no-useless-non-greedy`.

Then let's rename `no-useless-non-greedy` to `no-useless-lazy` before releasing v0.8.

Also, how do you plan to release new/change rules? We will add and change a lot of rules in the next few days (and weeks). Do you plan to make one big release after all is done or do you want to make some releases in between?

RunDevelopment (approved):

ota-meshi:
> how do you plan to release

I would like to make a minor release about once every two weeks. This week, when I have time, I'll merge some PRs and release 0.8.  
After finishing all the changes, I would like to make a major version upgrade including config changes.

ota-meshi:
I opened PR to rename the rule #110

RunDevelopment:
> I would like to make a minor release about once every two weeks.

Sounds like a good plan.

PR #101, opened 2021-04-10
Refactored optimal lookahead quantifier

RunDevelopment:
I did some refactoring to avoid code duplication.

I also noticed that the rule isn't correct. Some reported quantifiers might contain capturing groups in which case we can't change their quantifiers without changing the pattern.

I only added a TODO for now because this branch doesn't know `regexp-ast-analysis` yet. After master is merged into either this branch or #97, a check can be added using `hasSomeDescendant`.


ota-meshi:
Thank you!

PR #103, opened 2021-04-10
Fixed `match-any` false negatives and improved fixer

RunDevelopment:
- The rule now uses `matchesAllCharacters` for detection
- The fixer can now fix `dotAll` as well.
- New `parseFlags` function.


ota-meshi (approved):
It's amazing!

PR #105, opened 2021-04-10
Extracted common fix logic into functions

RunDevelopment:
I noticed that many fix functions share a lot of code, so I extracted the shared logic into two functions: `fixReplaceNode` and `fixReplaceQuant`.

The new functions could also be used in `match-any` but I haven't used them there yet to avoid conflicts with #103. I will add this after #103.


ota-meshi:
It's good, thank you!

PR #107, opened 2021-04-12
Added command starting the website to README

### LINKED ISSUE #100: Start website locally
Is there an easy way to start the website locally? There doesn't seem to be a script that does it.

It's also quite difficult to just start a local server because all paths in the website itself are prefixed with "/eslint-plugin-regexp/". Removing that prefix will cause Vue to generate different _absolute_ paths. This enables me that start the website locally but it will break the live website.

RunDevelopment:
This resolves #100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ module.exports = {
 
 This plugin provides one config:
 
-- `plugin:regexp/recommended` ... This is the recommended configuration for this plugin.  
+- `plugin:regexp/recommended` ... This is the recommended configuration for this plugin.
   See [lib/configs/recommended.ts](https://github.com/ota-meshi/eslint-plugin-regexp/blob/master/lib/configs/recommended.ts) for details.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,7 +67,7 @@ module.exports = {
 
 This plugin provides one config:
 
-- `plugin:regexp/recommended` ... This is the recommended configuration for this plugin.
```
RunDevelopment, line 70: My editor automatically trims trailing spaces. Should I revert these changes?

ota-meshi: If we want to break a line, you need two spaces, but I think we don't have to break a line here 😄

ota-meshi (approved):
Thank you!

README.md
RunDevelopment: Is that a common Markdown feature or specific to GH-flavor Markdown? 

I always use `<br>` for line breaks btw. It's more explicit.

PR #111, opened 2021-04-13
`no-useless-backreference`: Made compatible with `clean-regex/no-empty-backreference`

RunDevelopment:
- The now behaves the same as `clean-regex/no-empty-backreference`:
  - It will now also report backreferences to capturing groups that do not consume characters (e.g. `/(\b)\1/`).
- Error reports now use a better location property: 
   image
- The implementation now takes advantage of `regexp-ast-analysis` which makes the implementation simpler.

**TODO:**

~~The documentation has to be updated because now we do a little more than ESLint's `no-useless-backreference` rule.~~

---

@ota-meshi Reporting back refs with zero-length capturing groups is the only functional change to this rule. Is this a breaking change? Should I make a new rule for this instead?


ota-meshi:

I think it's good to change the `regexp/no-useless-backreference` rule.
It's a good to document that it can find more problems than ESLint's `no-useless-backreference` rule.

>  Is this a breaking change?

I think it's a feature addition that isn't a breaking change.

RunDevelopment:
I change the recommended config to prefer this rule. It will now disable the ESLint rule (to prevent double reporting) and enabled this rule. I also set its `recommended` property to `true`. 

We just talked about this being breaking but I don't if this also applies here.

ota-meshi:
> We just talked about this being breaking but I don't if this also applies here.

I don't think this is a breaking change.

RunDevelopment:
Perfect! In that case, I'll just add a bit more doc and mark it as ready for review after.

ota-meshi (approved):
Thank you for a lot of work! LGTM!

PR #112, opened 2021-04-13
Consistent message formatting

### LINKED ISSUE #106: Consistent formatting in messages
I noticed that regex expressions are quoted inconsistently. Sometime [single quotes](https://github.com/ota-meshi/eslint-plugin-regexp/blob/9b0a1f93edae35ae388e93b0eaa0bdcf68c389a6/lib/rules/negation.ts#L26) are used and sometimes [double quotes](https://github.com/ota-meshi/eslint-plugin-regexp/blob/9b0a1f93edae35ae388e93b0eaa0bdcf68c389a6/lib/rules/no-dupe-characters-character-class.ts#L261) are used. We should use one quoting style consistently instead.

([This rule](https://github.com/ota-meshi/eslint-plugin-regexp/blob/9b0a1f93edae35ae388e93b0eaa0bdcf68c389a6/lib/rules/prefer-regexp-test.ts#L23) even uses backticks but not for regexes.)

Which quoting style do we want? Single, double, or backticks?

RunDevelopment:
This resolves #106.


ota-meshi (approved):
Thank you!

Issue #91, opened 2021-04-02
Add rules that "eslint-plugin-clean-regex" has


ota-meshi:
Check each rule that [eslint-plugin-clean-regex](https://github.com/RunDevelopment/eslint-plugin-clean-regex) has, add the missing rule, and if eslint-plugin-regexp already has a similar rule, check for missing validation is needed.
Also, if we keep a note of the rule correspondence, it may be a document when creating a migration tool.

If we want to discuss one rule, we can also open a new issue.

### Problems

- [x] [clean-regex/confusing-quantifier](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/confusing-quantifier.md) ... Warn about confusing quantifiers.  
  - #93 Add new rule `regexp/confusing-quantifier`.
- [x] [clean-regex/disjoint-alternatives](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/disjoint-alternatives.md) ... Disallow different alternatives that can match the same words.
  - Handled by #179. 
- [x] [clean-regex/no-empty-alternative](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/no-empty-alternative.md) ... Disallow alternatives without elements.
  - #95 Add new rule `regexp/no-empty-alternative`.
- [x] [clean-regex/no-empty-backreference](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/no-empty-backreference.md) ... Disallow backreferences that will always be replaced with the empty string.
  - #111 makes `no-useless-backreference` functionally equivalent to `clean-regex/no-empty-backreference`.
- [x] [clean-regex/no-empty-lookaround](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/no-empty-lookaround.md) ... Disallow lookarounds that can match the empty string.
  - Similar to `regexp/no-empty-lookarounds-assertion`, but it doesn't catch potential empty assertions. `regexp/no-empty-lookarounds-assertion` needs to be changed. (#113)
- [x] [clean-regex/no-lazy-ends](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/no-lazy-ends.md) ... Disallow lazy quantifiers at the end of an expression.
  - #96 Add new rule `regexp/no-lazy-ends`.
[...]

PR #120, opened 2021-04-14
Added `regexp/no-trivially-nested-assertion` rule

RunDevelopment:
This is the `clean-regex/no-trivially-nested-lookaround` rule but with a different name and [this bug](https://github.com/ota-meshi/eslint-plugin-regexp/issues/91#issuecomment-817773291) fixed.

I renamed it because the "trivially nested assertion" is the inner assertion (e.g. in `/(?=(?!a))/`, the trivially nested assertion is `(?!a)`). Since the inner assertion can be any assertion and not just a lookaround, I renamed the rule.

### DIFF
--- a/lib/rules/no-trivially-nested-assertion.ts
+++ b/lib/rules/no-trivially-nested-assertion.ts
@@ -62,7 +62,9 @@ export default createRule("no-trivially-nested-assertion", {
     meta: {
         docs: {
             description: "disallow trivially nested assertions",
-            recommended: true,
+            // TODO Switch to recommended in the major version.
+            // recommended: true,
+            recommended: false,
         },
         fixable: "code",
         schema: [],
--- a/lib/rules/no-trivially-nested-assertion.ts
+++ b/lib/rules/no-trivially-nested-assertion.ts
@@ -1,0 +1,139 @@
+import type { Expression } from "estree"
+import type { RegExpVisitor } from "regexpp/visitor"
+import type {
+    Node as RegExpNode,
+    Assertion,
+    LookaroundAssertion,
+} from "regexpp/ast"
+import {
+    createRule,
+    defineRegexpVisitor,
+    fixReplaceNode,
+    getRegexpLocation,
+} from "../utils"
+import { hasSomeDescendant } from "regexp-ast-analysis"
+
+/**
+ * Returns whether the given node is a lookaround.
+ */
+function isLookaround(node: RegExpNode): node is LookaroundAssertion {
+    return (
+        node.type === "Assertion" &&
+        (node.kind === "lookahead" || node.kind === "lookbehind")
+    )
+}
+
+/**
+ * If the given lookaround only contains a single assertion, then this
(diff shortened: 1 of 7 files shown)


lib/rules/no-trivially-nested-assertion.ts
```diff
@@ -0,0 +1,139 @@
...
+        return `(?<${assertion.negate ? "=" : "!"}${assertion.raw.slice(4)}`
+    }
+    return null
+}
+
+export default createRule("no-trivially-nested-assertion", {
+    meta: {
+        docs: {
+            description: "disallow trivially nested assertions",
+            recommended: true,
```
ota-meshi, line 65:
```suggestion
            // TODO Switch to recommended in the major version.
            // recommended: true,
            recommended: false,
```

ota-meshi (changes requested):
Thank you for this PR! Almost LGTM.

Is it better to change the `new` script?

ota-meshi (approved):
Thank you for changing it!

RunDevelopment:
@ota-meshi Do you have a merge policy? I just merge past PRs as soon as they were approved and mergeable to move quickly. Should I not have done that?

ota-meshi:
I was wondering when to merge 😅.
You and I are the two maintainers of this plugin, so do you think it would be good to merge them as soon as we `Approve`?
If so, I will merge as soon as I think your PR is good.

RunDevelopment:
> You and I are the two maintainers of this plugin

No, I'm currently only a collaborator. This project isn't in an organization, so there can't be any maintainers from my understanding.

> do you think it would be good to merge them as soon as we Approve?

I think it is. No need to keep PRs open if you think they are good to go.

If the author of a PR thought that the PR still needs some changes, then the author should either say that or convert the PR into a draft.

---

I'll merge this now since you approved :)

ota-meshi:
> No, I'm currently only a collaborator. This project isn't in an organization, so there can't be any maintainers from my understanding.

I didn't know much about it. Do you think you should move to an organization for this project?  
(I'm not familiar with how to create an GitHub organization yet.)

RunDevelopment:
> Do you think this project should move to an organization for this project?

AFAIK, the main reasons people make organizations are:

1. Organization multiple projects.
1. Organizing teams/people. (This is what makes [maintainers](https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/giving-team-maintainer-permissions-to-an-organization-member).)
1. Moneatziation/Sponsioring.

We don't have multiple projects or sponsoring right now, so that's not an issue. If more people joined or you wanted to someday stop working on the project, then we can make an organization. Until then, I don't think we need an organization.

The only difference between maintainers and collaborators is that collaborators can't access the settings tab. So it's not really a hindrance, right now. 

That being said, I very much appreciate your offer.

ota-meshi:
Thank you for sharing your thoughts.

> Until then, I don't think we need an organization.

I understand. If we think we need something, let us talk again.

PR #125, opened 2021-04-14
Fix wrong autofix for non-RegExp arguments in `regexp/prefer-regexp-test` rule

### LINKED ISSUE #119: `prefer-regexp-test` doesn't account for non-RegExp arguments
The `String.prototype.match` function [allows non-RegExp arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#a_non-regexp_object_as_the_parameter) which isn't accounted for. This causes incorrect auto fixes.

For more information, see sindresorhus/eslint-plugin-unicorn#1150.

ota-meshi:
Fixed #119

### DIFF
--- a/lib/rules/prefer-regexp-test.ts
+++ b/lib/rules/prefer-regexp-test.ts
@@ -43,6 +43,7 @@ export default createRule("prefer-regexp-test", {
                     }
                     const arg = node.arguments[0]
                     const evaluated = getStaticValue(arg, context.getScope())
+                    let argIsRegExp = true
                     if (evaluated && evaluated.value instanceof RegExp) {
                         if (evaluated.value.flags.includes("g")) {
                             return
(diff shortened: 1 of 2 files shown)


lib/rules/prefer-regexp-test.ts
```diff
@@ -43,11 +43,14 @@ export default createRule("prefer-regexp-test", {
...
-                    ) {
+                    if (evaluated && evaluated.value instanceof RegExp) {
+                        if (evaluated.value.flags.includes("g")) {
+                            return
+                        }
+                    } else if (!typeTracer.isRegExp(arg)) {
+                        // Not RegExp
+                        // String.prototype.match function allows non-RegExp arguments
+                        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#a_non-regexp_object_as_the_parameter
                         return
```
fisker, line 54: Personally, I'd prefer still report on this case, but remove the auto-fix. 

Because it's just `foo.match(string)` -> `new RegExp(string).test(foo)`, `RegExp#test()` should still suggestted.

ota-meshi: Thank you for review! I think your idea is very good!
I will change it.

RunDevelopment (approved):

PR #126, opened 2021-04-15
Fix wrong autofix in `regexp/no-useless-range` rule

### LINKED ISSUE #104: `no-useless-range` fail to escape some characters
Here are some character classes that will be fixed incorrectly:

```js
/[,--b]/ -> /[,-b]/
/[\c-d]/ -> /[\cd]/
```

These examples are very artificial but should be fixed at some point. This is a low-priority bug.

ota-meshi:
Fixes #104


RunDevelopment (changes requested):
See #130.

RunDevelopment (approved):
LGTM

PR #130, opened 2021-04-15
Fixed escaping

RunDevelopment:
The fix in #126 was still wrong for some edge cases.

The problem we are trying to solve is that some characters get re-interpreted. I.e. in `[A-B-C]`, transforming it to `[AB-C]` will cause the second `-` to become a range operator. We are dealing with a parser problem.

To fix this problem, I only look at the raw source code around the range I'm trying to fix. This makes detect these cases where characters get interpreted as something else way easier. The general idea is still the same as in #126.


ota-meshi (approved):
Wonderful! Thank you.

PR #131, opened 2021-04-15
Added suggestions for `no-octal`

### LINKED ISSUE #115: `no-octal`: Provide suggestions
Maybe we could [provide a suggestion](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) to replace an octal escape with a hexadecimal escape (e.g. `\04` -> `\x04`)?

We can't apply fixes automatically for this rule but we can suggest a fix the user can apply themself.

RunDevelopment:
This resolves #115.


ota-meshi (changes requested):
You can use `suggestions` to test the suggestions that the rule provides.
See #134.

RunDevelopment:
Thanks for #134! I completely forgot to add tests...

ota-meshi (approved):
LGTM!

PR #132, opened 2021-04-15
Use `npm ci` instead of `npm install`

RunDevelopment:
Since we now have a package-lock, we might as well use it :)

This replaces all CI commands using `npm install` with [`npm ci`](https://docs.npmjs.com/cli/v7/commands/npm-ci/). This will guarantee that the CI installs the package specified in package-lock and is significantly faster than `npm install`.


ota-meshi (approved):
You're right! Thank you!

PR #135, opened 2021-04-16
Add `regexp/no-optional-assertion` rule

ota-meshi:
Changed to avoid some recursive processing compared to [clean-regex/no-optional-assertion](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/lib/rules/no-optional-assertion.ts).

### DIFF
--- a/docs/rules/no-obscure-range.md
+++ b/docs/rules/no-obscure-range.md
@@ -59,7 +59,7 @@ It allows all values that the [allowedCharacterRanges] setting allows.
 
 ### `"allowed": "alphanumeric"`
 
-<eslint-code-block fix>
+<eslint-code-block>
 
 ```js
 /* eslint regexp/no-obscure-range: ["error", { "allowed": "alphanumeric" }] */
@@ -80,7 +80,7 @@ var foo = /[😀-😄]/u;
 
 ### `"allowed": "all"`
 
-<eslint-code-block fix>
+<eslint-code-block>
 
 ```js
 /* eslint regexp/no-obscure-range: ["error", { "allowed": "all" }] */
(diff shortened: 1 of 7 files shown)


docs/rules/no-optional-assertion.md
```diff
@@ -0,0 +1,65 @@
...
+
+More generally, an assertion is optional, if the concatenation of all possible
+paths that start at the start of a zero-quantified element, end at the end of
+that element, and contain the assertion does not consume characters.
+
+Here's an example of that: `a(?:foo|(?<!-)(?:-|\b))*b`. The `\b` is optional.
+The lookbehind is not optional because following group can consume a character.
+
+The presence of optional assertions don't change the meaning of the pattern, so
+they are dead code.
```
RunDevelopment, line 31:
```suggestion
Assertions that are quantified (directly or indirectly) can be considered optional if the quantifier has a minimum of zero.

A simple example is the following pattern: `/a(?:$)*b/`. The `$` assertion will reject but if that happens, it will simply be ignored because of the `*` quantifier. The assertion is optional, serving no function whatsoever.

More generally, an assertion is optional, if there exists a parent quantifier with a minimum of zero such that all possible paths of the quantified element that contain the assertion do not consume characters.

Here's an example: `a(?:foo|(?<!-)(?:-|\b))*b`. The `\b` is optional. However, the lookbehind `(?<!-)` is not optional because the group `(?:-|\b)` right after it can consume a character.

Optional assertions don't affect the pattern in any way. They are essentially dead code.
```

RunDevelopment (changes requested):
LGTM apart from the docs... Sorry.

ota-meshi:
Thank you for your review! I made changes to this PR.

docs/rules/no-obscure-range.md
```diff
@@ -59,7 +59,7 @@ It allows all values that the [allowedCharacterRanges] setting allows.
 
 ### `"allowed": "alphanumeric"`
 
-<eslint-code-block fix>
```
RunDevelopment, line 62: Ah, my bad. Sorry.

RunDevelopment (approved):

docs/rules/no-obscure-range.md
ota-meshi: No problem 😄.

PR #136, opened 2021-04-16
Add `regexp/no-potentially-useless-backreference` rule

ota-meshi:


lib/rules/no-potentially-empty-backreference.ts
```diff
@@ -0,0 +1,60 @@
...
+                "disallow backreferences that reference a group that might not be matched",
+            // TODO Switch to recommended in the major version.
+            // recommended: true,
+            recommended: false,
+            default: "warn",
+        },
+        schema: [],
+        messages: {
+            potentiallyEmptyBackreference:
+                "Some path leading to the backreference do not go through the referenced capturing group without resetting its text.",
```
RunDevelopment, line 22:
```suggestion
                "Some paths leading to the backreference do not go through the referenced capturing group or the captured text might be reset before reaching the backreference.",
```

tests/lib/rules/no-potentially-empty-backreference.ts
```diff
@@ -0,0 +1,92 @@
...
+
+        // done by regexp/no-useless-backreference
+        String.raw`/(a+)b|\1/`,
+    ],
+    invalid: [
+        {
+            code: String.raw`
+            /* ✓ GOOD */
+            var foo = /(a+)b\1/;
+            var foo = /(a+)b|\1/;  // this will be done by regexp/no-useless-backreference
```
RunDevelopment, line 26: We already test this regex.

docs/rules/no-potentially-empty-backreference.md
```diff
@@ -0,0 +1,61 @@
...
+
+## :book: Rule Details
+
+If the referenced group of a backreference is not matched because some other
+path leads to the backreference, the backreference will be replaced with the
+empty string. The same will happen if the captured text of the referenced group
+was reset before reaching the backreference.
+
+This will handle backreferences which will always be replaced with the empty
+string for the above reason. Use [regexp/no-useless-backreference] for that.
```
RunDevelopment, line 21:
```suggestion
If the referenced group of a backreference is not matched because some other path leads to the backreference, the backreference will trivially accept (e.g. `/(?:(a)|b)\1/`). The same will happen if the captured text of the referenced group was reset before reaching the backreference.

This will not handle backreferences that always trivially accept. Use [regexp/no-useless-backreference] for that.
```

RunDevelopment (changes requested):
I must have really been sleep-deprived when I wrote the docs for this rule...

I left you a few suggestions to fix my mistakes but I couldn't think of a better description. The current rule description doesn't really convey what the rule is about IMO.

Also, I think we need a different name for this rule. I named `no-potentially-empty-backreference` because there was the `no-empty-backreference` rule. However, we have a `no-useless-backreference` rule instead. But I don't think that the name ``no-potentially-useless-backreference`` would be good either...

ota-meshi:
Thank you for your review!
I changed the rule name and made some changes.

I don't think the descriptions in the documentation is confusing to the user. Even if the user does not understand the description, I think that the user will understand it because there is example code.

RunDevelopment (approved):

PR #137, opened 2021-04-16
Add `regexp/no-useless-assertions` rule

ota-meshi:
This PR adds `regexp/no-useless-assertions` rule.
The `regexp/no-useless-assertions` rule that is the same rule as the [clean-regex/no-unnecessary-assertions](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/no-unnecessary-assertions.md) rule.
This plugin has used "useless" as the meaning of "unnecessary" in the existing rule names, so I renamed the new rule.

### DIFF
--- a/lib/rules/no-useless-assertions.ts
+++ b/lib/rules/no-useless-assertions.ts
@@ -21,7 +21,6 @@ import {
     hasSomeDescendant,
     isPotentiallyEmpty,
 } from "regexp-ast-analysis"
-import { JS } from "refa"
 
 const messages = {
     alwaysRejectByChar:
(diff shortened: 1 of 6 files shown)


lib/rules/no-useless-assertions.ts
```diff
@@ -0,0 +1,319 @@
...
+                        })
+                    } else {
+                        // only if the character is a sub set of /./, will the assertion trivially reject
+
+                        // with this little flag hack, we can easily create the dot set.
+
+                        const dot = JS.createCharSet(
+                            [{ kind: "any" }],
+                            flagsWithoutDotAll,
+                        )
```
RunDevelopment, line 124: You can use the `Chars` namespace from `regexp-ast-analysis` here.

```suggestion
                        const dot = Chars.lineTerminator(flags).negate()
```

RunDevelopment (changes requested):
One minor nit.

ota-meshi:
Thank you for letting me know how to do that!
I made changes to this PR.

RunDevelopment (approved):

PR #139, opened 2021-04-16
Add `regexp/no-useless-flag` rule

### LINKED ISSUE #118: Rule proposal: Forbid `g` flag used on `RegExp#test()`
If the regex is only used once on `RegExp#test()`, the `g` flag is useless.

#### Fail

```js
/foo/g.test(bar);
```

```js
new RegExp('foo', 'g').test(bar);
```

#### Pass

```js
const regex = /foo/g;
regex.test(bar);
```

I original make this proposal [here](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1174), but feel better put `RegExp` related rules together in this codebase.

ota-meshi:
This PR adds `regexp/no-useless-flag` rule.
The `regexp/no-useless-flag` rule reports unnecessary flags.

The `i`, `m` and `s` flag checks were merged from [clean-regex/no-unnecessary-flag].  
The `g` flag checks was suggested in #118.

close #118

[clean-regex/no-unnecessary-flag]: https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/no-unnecessary-flag.md

### DIFF
--- a/docs/rules/no-useless-flag.md
+++ b/docs/rules/no-useless-flag.md
@@ -18,7 +18,7 @@ This will point out present regex flags that do not change the pattern.
 ### `i` flag (ignoreCase)
 
 The `i` flag is only necessary if the pattern contains any characters with case
-variations. If no such characters are part of the pattern, the flag is
+variations. If the pattern contains no such characters, the flag will be
 unnecessary. E.g. `/\.{3}/i`
 
 <eslint-code-block fix>
@@ -38,8 +38,8 @@ var foo = /\w+/i;
 
 ### `m` flag (multiline)
 
-The `m` flag changes the meaning of the `^` and `$` anchors, so if the pattern
-doesn't contain these anchors, it's unnecessary. E.g. `/foo|[^\r\n]*/m`
+The `m` flag changes the behavior of the `^` and `$` assertions. If the pattern
+doesn't contain these anchors, the `m` flag will be unnecessary. E.g. `/foo|[^\r\n]*/m`
 
 <eslint-code-block fix>
 
@@ -59,8 +59,8 @@ var foo = /a|b/m;
 ### `s` flag (dotAll)
 
 The `s` flag makes the dot (`.`) match all characters instead of the usually
-non-line-terminator characters, so if the pattern doesn't contain a dot
-character set, it will be unnecessary. E.g. `/[.:]/s`
+non-line-terminator characters. If the pattern doesn't contain a dot
+character set, the `s` flag will be unnecessary. E.g. `/[.:]/s`
 
 <eslint-code-block fix>
 
--- a/docs/rules/no-useless-flag.md
+++ b/docs/rules/no-useless-flag.md
@@ -2,20 +2,86 @@
 pageClass: "rule-details"
 sidebarDepth: 0
 title: "regexp/no-useless-flag"
-description: "disallow unused global flag"
+description: "disallow unnecessary regex flags"
 ---
 # regexp/no-useless-flag
 
-> disallow unused global flag
+> disallow unnecessary regex flags
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 
 ## :book: Rule Details
 
-This rule checks if a regular expression with the `g` flag is used in the global regular expression test.
-If it is determined that the global flag is unnecessary, it will be reported.
+This will point out present regex flags that do not change the pattern.
 
-<eslint-code-block>
+### `i` flag (ignoreCase)
+
+The `i` flag is only necessary if the pattern contains any characters with case
+variations. If no such characters are part of the pattern, the flag is
+unnecessary. E.g. `/\.{3}/i`
+
+<eslint-code-block fix>
+
+```js
+/* eslint regexp/no-useless-flag: "error" */
+
+/* ✓ GOOD */
+var foo = /a|b/i;
+
+/* ✗ BAD */
+var foo = /\.{3}/i;
+var foo = /\w+/i;
+```
+
+</eslint-code-block>
+
+### `m` flag (multiline)
+
+The `m` flag changes the meaning of the `^` and `$` anchors, so if the pattern
+doesn't contain these anchors, it's unnecessary. E.g. `/foo|[^\r\n]*/m`
+
+<eslint-code-block fix>
+
+```js
+/* eslint regexp/no-useless-flag: "error" */
+
+/* ✓ GOOD */
+var foo = /^foo$/m;
+
+/* ✗ BAD */
+var foo = /foo|[^\r\n]*/m;
+var foo = /a|b/m;
+```
+
+</eslint-code-block>
+
+### `s` flag (dotAll)
+
+The `s` flag makes the dot (`.`) match all characters instead of the usually
+non-line-terminator characters, so if the pattern doesn't contain a dot
+character set, it will be unnecessary. E.g. `/[.:]/s`
+
+<eslint-code-block fix>
+
+```js
+/* eslint regexp/no-useless-flag: "error" */
+
+/* ✓ GOOD */
+var foo = /a.*?b/s;
+
+/* ✗ BAD */
+var foo = /[.:]/s;
+var foo = /^foo$/s;
+```
+
+</eslint-code-block>
+
+### `g` flag (global)
+
+The `g` flag is used when you need to test a regular expression against all possible string match. If not, it will be unnecessary.
+
+<eslint-code-block fix>
 
 ```js
 /* eslint regexp/no-useless-flag: "error" */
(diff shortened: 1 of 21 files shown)


RunDevelopment:
We will soon bring over the [`clean-regex/no-unnecessary-flag`](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/master/docs/rules/no-unnecessary-flag.md) rule. Wouldn't it make sense to add merge the two rules?

ota-meshi:
I haven't checked the `clean-regex/no-unnecessary-flag` rule until now.
I think these rules can be merged into one rule.

ota-meshi:
I would like to change this PR to merge `clean-regex/no-unnecessary-flag`. (I also change the rule name.)

ota-meshi:
I merged the clean-regex/no-unnecessary-flag rule.

docs/rules/no-useless-flag.md
```diff
@@ -0,0 +1,164 @@
...
+var foo = /\.{3}/i;
+var foo = /\w+/i;
+```
+
+</eslint-code-block>
+
+### `m` flag (multiline)
+
+The `m` flag changes the meaning of the `^` and `$` anchors, so if the pattern
+doesn't contain these anchors, it's unnecessary. E.g. `/foo|[^\r\n]*/m`
```
RunDevelopment, line 42:
```suggestion
The `m` flag changes the behavior of the `^` and `$` assertions. If the pattern
doesn't contain these anchors, the `m` flag will be unnecessary. E.g. `/foo|[^\r\n]*/m`
```

docs/rules/no-useless-flag.md
```diff
@@ -0,0 +1,164 @@
...
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This will point out present regex flags that do not change the pattern.
+
+### `i` flag (ignoreCase)
+
+The `i` flag is only necessary if the pattern contains any characters with case
+variations. If no such characters are part of the pattern, the flag is
```
RunDevelopment, line 21:
```suggestion
variations. If the pattern contains no such characters, the flag will be
```

docs/rules/no-useless-flag.md
```diff
@@ -0,0 +1,164 @@
...
+var foo = /a|b/m;
+```
+
+</eslint-code-block>
+
+### `s` flag (dotAll)
+
+The `s` flag makes the dot (`.`) match all characters instead of the usually
+non-line-terminator characters, so if the pattern doesn't contain a dot
+character set, it will be unnecessary. E.g. `/[.:]/s`
```
RunDevelopment, line 63:
```suggestion
non-line-terminator characters. If the pattern doesn't contain a dot
character set, the `s` flag will be unnecessary. E.g. `/[.:]/s`
```

RunDevelopment (changes requested):
See #145.

Apart from that, I mainly just want to fix my English...

RunDevelopment:
I forgot. We need to talk about fix conflicts here.

Right now, the fixes only replace a single flag. This is a problem because that means that ESLint is allowed to apply other fixes to the regex _simultaneously_.

Consider the regex `/[^]/s`. The `no-useless-flag` rule will try to remove the `s` flag but the `match-any` rule will try to use the `s` and transform `[^]` -> `.`. Since the fixes for both rules change different parts of the regex, ESLint will apply them both and we get the regex `/./`. Not what we wanted. 
(I used the `s` flag as an example but all other flags have the same problem in one form or another.)

To fix this, we have to increase the range of fixes to include their _dependencies_ as well. E.g. the fix `[^]` -> `.` depends on the `s` flag, so it has to be included in the fix as well. 

In `clean-regex`, I will [replace the whole regex](https://github.com/RunDevelopment/eslint-plugin-clean-regex/blob/a41e31c6d3dd10650516512187a03af423a4e495/lib/rules-util.ts#L111) if a fix depends on any flag. Fixed that change flags and depend on elements will similarly replace the whole regex as well. This solves the issue because two fixes that depend on/change flags will now _conflict_ which means that ESLint can only apply one fix.

However, this might be difficult to implement here because we support a little more than just regexp literals.

We will also have to carefully go through all existing rules and add all dependencies to existing fixes. 

(Rules that don't provide fixes are not affected. Even suggestions are fine.)

lib/rules/no-useless-flag.ts
```diff
@@ -0,0 +1,598 @@
...
+                visitor,
+                createUselessGlobalFlagVisitor(context),
+            )
+        }
+        return visitor
+    },
+})
+
+/** Get a parent node */
+function getParent<E extends Node>(node: Node | null): E | null {
```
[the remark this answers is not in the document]
ota-meshi: Thank you for letting me know this. Other rules have the same function, so I was thinking of making it common, but I forgot 😓

ota-meshi:
Thanks for pointing out the issue of autofix conflicts!

I changed the autofix range of the `no-useless-flag` and `match-any` rules to avoid conflicts.
I've seen other rules that use flags, but I didn't think they would cause a conflict. So I haven't changed any other rules.

lib/rules/no-useless-flag.ts
```diff
@@ -0,0 +1,631 @@
...
+) {
+    if (node.type === "Literal") {
+        const flagIndex =
+            node.range![1] -
+            node.regex.flags.length +
+            node.regex.flags.indexOf(flag)
+        return [
+            // Replace the range of regular expression literals to avoid conflicts.
+            fixer.replaceTextRange([node.range![0], node.range![0] + 1], "/"),
+            fixer.removeRange([flagIndex, flagIndex + 1]),
```
RunDevelopment, line 215: Could you please explain why this works?

From what I see, ESLint is free to apply the fixes we give independently of each other. Does ESLint guarantee that an array of fixes will always be applied atomically?

ota-meshi: Even if multiple ranges are given by the fixer of one report, ESLint synthesizes them and judges it as one fix. If the fix range returned by each report is duplicated, only one will be applied.

Regular expressions start with `/`, so by replacing the start position with `/`, the fix range of this report will be from the start position of the regular expression.

ota-meshi: https://github.com/eslint/eslint/blob/master/lib/linter/report-translator.js#L146

ota-meshi: The maximum number of fixes per report is one, and multiple reports are applied to the fix range without conflict in the following code:
https://github.com/eslint/eslint/blob/master/lib/linter/source-code-fixer.js#L61

RunDevelopment: The problem that I have with this is that [ESLint doesn't guarantee this behavior](https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes). ESLint is free to change this part of its implementation at any time.

I think we should not rely on this specific implementation detail.

ota-meshi: I'll explain this, but through this discussion I realized that this code is hard to read and requires a lot of information. I would like to change this source code.

---

I think you are worried about the possibility that the merge process (`mergeFixes`) will be removed.

I think removing `mergeFixes` involves changing the API of `LintMessage#fix`, so removing it should be a breaking change.

https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type
https://eslint.org/docs/developer-guide/nodejs-api#-editinfo-type

Next is why `mergeFixes` is guaranteed to be done before applying the fix (`SourceCodeFixer.applyFixes`).

The `LintMessage` API can be used in the `postprocess` of `processors` in the eslint-plugin.

https://eslint.org/docs/developer-guide/working-with-plugins#processors-in-plugins

The `postprocess` is done before the `SourceCodeFixer.applyFixes`, so you can shift the fix locations. This is often done with plugins such as `eslint-plugin-markdown`, `eslint-plugin-html` and `eslint-plugin-svelte3`.

Therefore, removing `mergeFixes` is a breaking change and will not be done except for major version upgrades.

RunDevelopment: > I think you are worried about the possibility that the merge process (mergeFixes) will be removed.
>
> I think removing mergeFixes involves changing the API of LintMessage#fix, so removing it should be a breaking change.
>
> https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type
> https://eslint.org/docs/developer-guide/nodejs-api#-editinfo-type

That was exactly my worry. I was worried that one error with multiple fixes might be converted into many errors with one fix each.

But you are right, there is nothing to worry about. `LintMessage` provides the guarantees we need.

> The LintMessage value is the information of *each linting error*.

This is the important part. This means that every error report will result in exactly one `LintMessage`. `mergeFixes` is required to provide this guarantee.

tests/lib/rules/match-any.ts
```diff
@@ -110,14 +110,35 @@ tester.run("match-any", rule as any, {
         },
         {
             code: "/[\\s\\S] [\\S\\s] [^] ./s",
-            output: "/. . . ./s",
+            output: "/. [\\S\\s] [^] ./s",
```
RunDevelopment, line 113: I know that this is because of conflicts but it looks like a bug. Could you please add a comment here?

```suggestion
            // Only one character class gets fixed because they all depend on the `s` flag.
            // This shared dependency causes all of their fixes to conflict, so only one fix can be applied.
            output: "/. [\\S\\s] [^] ./s",
```

RunDevelopment (changes requested):

tests/lib/rules/match-any.ts
ota-meshi: Thank you!

RunDevelopment (approved):

PR #140, opened 2021-04-16
Added `prefer-named-backreference` rule

### LINKED ISSUE #128: New rule `prefer-named-backreferences`
Backreferences to named groups should probably refer to the group by name and not by number.

```js
// valid
- /(a)\1/u
- /(?<foo>a)\k<foo>/u

// invalid
- /(?<foo>a)\1/u
//          ^~
```

The rule should be auto-fixable.

RunDevelopment:
This resolves #128.


ota-meshi (approved):
LGTM!

PR #141, opened 2021-04-16
Fixed `update` script

RunDevelopment:
@ota-meshi You probably also noticed, that I broke the `update` script in #122.

This PR 1) fixes the update script so that it no longer removes the "Settings" section and 2) adds a CI job to verify that `npm run update` doesn't change any files.


ota-meshi:
I did something similar to you.
And, I accidentally pushed it directly to master. Sorry.
https://github.com/ota-meshi/eslint-plugin-regexp/commit/7c5908c2249d746cdb06a2e830ddc5e3cf849586

RunDevelopment:
It's okay. But I'll make another PR for the CI job.

PR #142, opened 2021-04-16
Improved `no-useless-assertions` rule

RunDevelopment:
Just some minor improvement that makes the code a little cleaner.


ota-meshi (approved):
Thank you!

PR #143, opened 2021-04-16
Updated `no-useless-non-greedy` rule

RunDevelopment:
This makes the `regexp/no-useless-non-greedy` rule functionally equivalent to the `clean-regex/no-unnecessary-lazy` rule.


ota-meshi (approved):
That's great improvement! Thank you!

PR #144, opened 2021-04-16
New CI job to check for changes after `update`

RunDevelopment:
As promised, here is the CI job from #141.

This job is quite useful because it means that we can never forget to run `npm run update` again.


ota-meshi (approved):
This is a very good idea! Tnak you!

PR #145, opened 2021-04-16
`no-useless-flag`: Fixed i flag conditions

RunDevelopment:
The condition for `i` flags wasn't correct:

- `\w` is case sensitive in Unicode mode.
- `\b` is case sensitive in Unicode mode.
- Unicode property sets (e.g. `\p{Lu}`) were not accounted for.
- The condition for characters `char.toLowerCase() !== char.toUpperCase()` isn't correct. It's close to the actual [canonicalization function](https://tc39.es/ecma262/#sec-runtime-semantics-canonicalize-ch) used in non-Unicode mode but not correct. The new condition properly accounts for Unicode mode as well.

This is completely my fault, so I fixed it.

### DIFF
--- a/lib/rules/no-useless-flag.ts
+++ b/lib/rules/no-useless-flag.ts
@@ -6,6 +6,7 @@ import {
     FLAG_GLOBAL,
     FLAG_IGNORECASE,
     FLAG_MULTILINE,
+    parseFlags,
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
Thank you!

PR #146, opened 2021-04-16
Added `no-trivially-nested-quantifier` rule

RunDevelopment:
This brings over the `clean-regex/no-trivially-nested-quantifier` rule.

### DIFF
--- a/lib/rules/no-trivially-nested-quantifier.ts
+++ b/lib/rules/no-trivially-nested-quantifier.ts
@@ -1,0 +1,282 @@
+import type { Expression } from "estree"
+import type { RegExpVisitor } from "regexpp/visitor"
+import type { Node as RegExpNode, Quantifier } from "regexpp/ast"
+import type { Quant } from "../utils"
+import {
+    fixReplaceNode,
+    fixReplaceQuant,
+    quantToString,
+    createRule,
+    defineRegexpVisitor,
+    getRegexpLocation,
+} from "../utils"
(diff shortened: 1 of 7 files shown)


lib/rules/no-trivially-nested-quantifier.ts
```diff
@@ -0,0 +1,282 @@
...
+            return {
+                onQuantifierEnter(qNode) {
+                    if (hasSomeAncestor(qNode, (a) => ignore.has(a))) {
+                        return
+                    }
+                    if (qNode.max === 0) {
+                        // this rule does not handle this case
+                        ignore.add(qNode)
+                        return
+                    }
```
ota-meshi, line 163: I don't think you need to check it ancestors if you use a Leave event.

```ts
            let ignore: Quantifier | null = null

            return {
                onQuantifierLeave(qNode) {
                    if (ignore === qNode) {
                        ignore = null
                    }
                },
                onQuantifierEnter(qNode) {
                    if (ignore) {
                        return
                    }
                    if (qNode.max === 0) {
                        // this rule does not handle this case
                        ignore = qNode
                        return
                    }
                    // ...
                }
```

RunDevelopment: I don't think we need the `ignore` set at all, now that I think about it. I believe that I added this to avoid conflicts. (This was before I understood how fix conflicts really worked.)

ota-meshi (approved):
LGTM!

Issue #92, opened 2021-04-02
Refactor using "regexp-ast-analysis"


ota-meshi:
[refa](https://github.com/RunDevelopment/refa) package seems to have many useful utilities for regular expressions.
I think we can remove a lot of code from eslint-plugin-regexp by refactoring with it.

ota-meshi:
Currently, eslint-plugin-regexp has rules that exclude unicode properties (`\p{...}`) from validation.  
(That's because I didn't know how to get the character set contained in the unicode property.)  
I think we can use refa to check the characters contained in unicode properties.

RunDevelopment:
I changed the title to reflect what actually has to be done. regexp-ast-analysis is a lot easier to use than refa for our purposes.

ota-meshi:
🎉

PR #148, opened 2021-04-17
Updated `regexp/no-dupe-characters-character-class` rule using `regexp-ast-analysis`.

ota-meshi:
This PR contains the following changes:

- Fix false negatives for `[\s\u180e]` in `regexp/no-dupe-characters-character-class` rule.  
  (The document (Japanese MDN) I checked, `\s` contained `\u180e`, which was incorrect.)
- Fixed false negatives for `i` flag  in `regexp/no-dupe-characters-character-class` rule.  
- Add supports for unicode property to `regexp/no-dupe-characters-character-class` rule.
- Add supports for negative escape character class to `regexp/no-dupe-characters-character-class` rule.
- Update `regexp/no-dupe-characters-character-class` rule to report that escape character class include escape character class.
- Improved reporting of `regexp/no-dupe-characters-character-class` rule when fully including elements.

related to #92

### DIFF
--- a/lib/rules/no-dupe-characters-character-class.ts
+++ b/lib/rules/no-dupe-characters-character-class.ts
@@ -45,20 +45,20 @@ function getRangesIntersection(a: CharRange, b: CharRange): CharRange | null {
  * Gets the two given character class range intersection.
  * @param a The first character class range.
  * @param b The second character class range.
- * @returns { CharRange | null } the two given character class range is intersect.
+ * @returns the two given character class range is intersect.
  */
 function getCharacterClassRangesIntersection(
     a: NodeAndCharSet<CharacterClassRange>,
     b: NodeAndCharSet<CharacterClassRange>,
 ) {
-    return getRangesIntersection(getMinimumRange(a), getMinimumRange(b))
+    return getRangesIntersection(a.charSet.ranges[0], b.charSet.ranges[0])
 }
 
 /**
  * Gets the two given character class range ans character set intersections.
  * @param range The character class range.
  * @param set The character set.
- * @returns { (number | CharRange)[] } the two given character class range ans character set intersections.
+ * @returns the two given character class range ans character set intersections.
  */
 function getCharacterClassRangeAndCharacterSetIntersections(
     range: NodeAndCharSet<CharacterClassRange>,
@@ -75,7 +75,7 @@ function getCharacterClassRangeAndCharacterSetIntersections(
     /**
      * Checks if the given code point is at the edge of the range.
      * @param codePoint The code point to check
-     * @returns {boolean} `true` if the given code point is at the edge of the range.
+     * @returns `true` if the given code point is at the edge of the range.
      */
     function isCodePointIsRangeEdge(codePoint: number) {
         return (
@@ -133,19 +133,6 @@ function getCharacterClassRangeAndCharacterSetIntersections(
 }
 
 /**
- * Build key from CharacterClassRange
- */
-function getMinimumRange(classRange: NodeAndCharSet<CharacterClassRange>) {
-    let range: CharRange | null = null
-    for (const r of classRange.charSet.ranges) {
-        if (!range || range.min > r.min) {
-            range = r
-        }
-    }
-    return range!
-}
-
-/**
  * Grouping the given character class elements.
  * @param elements The elements to grouping.
  */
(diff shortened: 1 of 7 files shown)


RunDevelopment (changes requested):
A few minor nits.

Also, the way you cache `CharSet`s is awkward. Everything has to be a `NodeAndCharSet` which makes the code more complex.

You probably do this because `toCharSet` is a somewhat expensive operation. Then how about we make a cheap `toCharSet` instead? Basically, how about we make a cached `toCharSet` with a simplified function signature of `(element: ToCharSetElement) => CharSet`?

lib/rules/no-dupe-characters-character-class.ts
```diff
@@ -69,51 +33,44 @@ function isCharacterInCharacterSet(
...
-    return [Math.max(a[0], b[0]), Math.min(a[1], b[1])]
+    return { min: Math.max(a.min, b.min), max: Math.min(a.max, b.max) }
 }
 
 /**
  * Gets the two given character class range intersection.
  * @param a The first character class range.
  * @param b The second character class range.
- * @returns { [number, number] | null } the two given character class range is intersect.
+ * @returns { CharRange | null } the two given character class range is intersect.
```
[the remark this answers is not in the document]
ota-meshi: We don't need the types in TSDoc so I'll remove them (I originally made this rule in js).

ota-meshi:
Thank you for your review. I changed this PR.

> You probably do this because toCharSet is a somewhat expensive operation. Then how about we make a cheap toCharSet instead? Basically, how about we make a cached toCharSet with a simplified function signature of (element: ToCharSetElement) => CharSet?

I added helper to cache the result of toCharSet.

https://github.com/ota-meshi/eslint-plugin-regexp/pull/148/files#diff-76051e819bd9ea2a35405d89d3fbf1831f45019a2b12d5e0a915668d1bd46462R204

This is supplied via the 5th argument of createVisitor.

https://github.com/ota-meshi/eslint-plugin-regexp/pull/148/files#diff-76051e819bd9ea2a35405d89d3fbf1831f45019a2b12d5e0a915668d1bd46462R51

The cached results are shared by all rules and can be processed efficiently.

I want to refactor because there are too many arguments for createVisitor, but I need to do it when there is no PR to add the rule.

https://github.com/ota-meshi/eslint-plugin-regexp/pull/148/files#diff-76051e819bd9ea2a35405d89d3fbf1831f45019a2b12d5e0a915668d1bd46462R46

lib/rules/no-dupe-characters-character-class.ts
```diff
@@ -179,7 +163,7 @@ function groupingElements(
             }
         } else if (e.type === "CharacterClassRange") {
             const nodeAndCharSet = { node: e, charSet }
-            const range = getMinimumRange(nodeAndCharSet)
+            const range = charSet.ranges[0]
             const key = String.fromCodePoint(range.min, range.max)
```
RunDevelopment: Unfortunately, this doesn't uniquely identify a range. It's almost correct and it was difficult to find an example but here it is:

- `/[A-Z]/i` creates the ranges: `{ min: 65, max: 90 }, { min: 97, max: 122 }`
- `/[a-\uFFFF]/i` creates the ranges: `{ min: 65, max: 90 }, { min: 97, max: 65535 }`

As you can see, `ranges[0]` is the same for both.

The problem is that `ranges[0]` only works when there is no canonicalization (ignore case). Since we cannot assume anything about the canonicalization function, we have to use the whole char set as the key.
```suggestion
            const key = flags.ignoreCase ? charSet.toString() : String.fromCodePoint(charSet.ranges[0].min, charSet.ranges[0].max)
```

RunDevelopment (changes requested):

lib/rules/no-dupe-characters-character-class.ts
```diff
@@ -8,137 +8,72 @@ import type {
...
 function getCharacterClassRangesIntersection(
     a: CharacterClassRange,
     b: CharacterClassRange,
+    { toCharSet }: RegExpContext,
 ) {
-    return getRangesIntersection(
-        [a.min.value, a.max.value],
-        [b.min.value, b.max.value],
-    )
+    return getRangesIntersection(toCharSet(a).ranges[0], toCharSet(b).ranges[0])
```
RunDevelopment, line 47: Same problem here. This doesn't work in ignore case mode.

lib/rules/no-dupe-characters-character-class.ts
ota-meshi: Thank you for letting me know about this!
I didn't understand the range correctly 😓. I would like to fix this issues.

RunDevelopment: No worries. It also took me quite some time to figure out that this doesn't work.

ota-meshi:
Thank you for your review!
I fixed a bug when using range with ignoreCase.

RunDevelopment (changes requested):
See #151.

ota-meshi:
> See #151.

Thank you!

RunDevelopment (approved):

PR #150, opened 2021-04-18
Refactor `defineRegexpVisitor`

### LINKED ISSUE #149: Refactoring `defineRegexpVisitor(context,{createVisitor})`
Change the argument of createVisitor to context only.

See https://github.com/ota-meshi/eslint-plugin-regexp/pull/148#issuecomment-821915380

We also need to consider whether we can add other helper functions. e.g. `getRegexpLocation()`

ota-meshi:
This PR refactors `defineRegexpVisitor` to provide some helpers for each rule. This allows we to omit function arguments and shorten we code.

More details on helpers can be found in the `RegExpHelpers` type.
https://github.com/ota-meshi/eslint-plugin-regexp/pull/150/files#diff-76051e819bd9ea2a35405d89d3fbf1831f45019a2b12d5e0a915668d1bd46462R28

Functions provided by helpers

- toCharSet
- getRegexpLocation
- fixerApplyEscape
- fixReplaceNode
- fixReplaceQuant
- getRegexpRange

---

close #149

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -1,13 +1,7 @@
 import type * as ESTree from "estree"
 import type { RuleListener, RuleModule, PartialRuleModule } from "../types"
 import type { RegExpVisitor } from "regexpp/visitor"
-import type {
-    Alternative,
-    Element,
-    Node,
-    Node as RegExpNode,
-    Quantifier,
-} from "regexpp/ast"
+import type { Alternative, Element, Node, Quantifier } from "regexpp/ast"
 import { RegExpParser, visitRegExpAST } from "regexpp"
 import {
     CALL,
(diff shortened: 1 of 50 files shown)


lib/utils/index.ts
```diff
@@ -18,21 +18,99 @@ import {
...
+    /**
+     * Creates SourceLocation from the given regexp node
+     * @see getRegexpLocation
+     *
+     * @param regexpNode The regexp node to report.
+     * @param offsets The report location offsets.
+     * @returns The SourceLocation
+     */
+    getRegexpLocation: (
+        regexpNode: RegExpNode,
```
RunDevelopment, line 43: Right now, we have `import type { Node, Node as RegExpNode } from "regexpp/ast"`. Could you please remove either `Node` or `Node as RegExpNode`?

RunDevelopment (commented):
I really like this. The new API is really convenient and easy to use. Great work!

lib/utils/index.ts
ota-meshi: Thank you for letting me know! I fixed it.

RunDevelopment (approved):

PR #151, opened 2021-04-18
Simplified implementation

RunDevelopment:
I changed the implementation to only use `CharSet` functions. This made the `getCharacterClassRangeAndCharacterClassRangeOrCharacterSetIntersections` unnecessary which removed about 100 lines of code.

Also, the overlap between ranges and character sets is now reported as one report (see the `/[\wA-_]/` test case).

---

@ota-meshi This is just an assumption on my part but did you not know how to convert a `CharSet` into something readable? If you have any questions regarding what refa can do, feel free to ask any time.

You can also read [the docs](https://rundevelopment.github.io/refa/docs/latest/). The main points of interest are probably [CharSet](https://rundevelopment.github.io/refa/docs/latest/classes/charset.html) and [toLiteral](https://rundevelopment.github.io/refa/docs/latest/modules/js.html#toliteral).

### DIFF
--- a/lib/rules/no-dupe-characters-character-class.ts
+++ b/lib/rules/no-dupe-characters-character-class.ts
@@ -11,130 +11,10 @@ import type {
     AnyCharacterSet,
 } from "regexpp/ast"
 import type { RegExpContext } from "../utils"
-import {
-    createRule,
-    defineRegexpVisitor,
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
> removed about 100 lines of code.

Very cool! Thank you!

PR #158, opened 2021-04-19
Updated `prefer-character-class` rule

RunDevelopment:
This reimplements the `prefer-character-class` using the `clean-regex/prefer-character-class` implementation.

The `clean-regex` implementation is a lot more aggressive and will combine basically anything into a character class that is possible to combine. It will even ignore alternative order if it's safe to do so (e.g. `a|bb|c|d` -> `[acd]|bb`).


ota-meshi (approved):
It's amazing! Thank you!

PR #161, opened 2021-04-20
Update `no-dupe-characters-character-class#charSetToReadableString()` to be shared by rules.

ota-meshi:
This change includes `no-invisible-character` rule message and output source change.


lib/utils/index.ts
```diff
@@ -750,6 +751,29 @@ export function quantToString(quant: Readonly<Quant>): string {
     return value
 }
 
+/**
+ * Returns a regexp literal text of the given char set.
+ */
+export function charSetToRegExpText(
```
RunDevelopment, line 757: I think this function should be renamed. I mean, what is "regexp text"? "Text" might refer to a whole regexp literal (e.g. `/foo/i`) or just its source (e.g. `foo`). We should replace "text" with "source".

However, we can be even more specific than "source". `JS.toLiteral` guarantees that it produces a literal representation of the given RE AST. Since we only give it a single character class, it is guaranteed to produce a single character, character class, or character set. (I should have mentioned that in the doc.)

```suggestion
export function toCharSetSource(
```

RunDevelopment (changes requested):
Good idea!

However, I think the new functions should be merged and renamed.

ota-meshi:
Thank you for your review! I renamed the function and merged the two functions.

RunDevelopment (approved):

PR #163, opened 2021-04-20
Add `regexp/hexadecimal-escape` rule

### LINKED ISSUE #162: New rule to enforce consistent hex escape
Characters that can use hexadecimal escape can use both hexadecimal escape and unicode escape.
I think it useful to have a new rule that enforces the consistent use of hexadecimal escapes.

Rule Name: `regexp/hexadecimal-escape`
Options:
- `"always"` ... Expects that the hexadecimal escape will be used whenever possible. This is default.
- `"never"` ... Ensures that no hexadecimal escape is used in any characters.

ota-meshi:
close #162

### DIFF
--- a/lib/rules/hexadecimal-escape.ts
+++ b/lib/rules/hexadecimal-escape.ts
@@ -1,0 +1,125 @@
+import type { RegExpVisitor } from "regexpp/visitor"
+import type { Character } from "regexpp/ast"
+import type { RegExpContext } from "../utils"
+import {
+    defineRegexpVisitor,
+    createRule,
+    getEscapeSequenceKind,
+    EscapeSequenceKind,
+    toCharSetSource,
+    isHexadecimalEscape,
+} from "../utils"
+
+export default createRule("hexadecimal-escape", {
+    meta: {
+        docs: {
+            description: "enforce consistent usage of hexadecimal escape",
+            recommended: false,
+        },
+        fixable: "code",
+        schema: [
+            {
+                enum: ["always", "never"], // default always
+            },
+        ],
+        messages: {
+            expectedHexEscape:
+                "Expected hexadecimal escape ('{{hexEscape}}'), but {{unexpectedKind}} escape ('{{rejectEscape}}') is used.",
+            unexpectedHexEscape:
+                "Unexpected hexadecimal escape ('{{hexEscape}}').",
+        },
+        type: "suggestion", // "problem",
+    },
+    create(context) {
+        const always = context.options[0] !== "never"
+
+        /**
+         * Verify for always
+         */
+        function verifyForAlways(
+            { node, getRegexpLocation, fixReplaceNode }: RegExpContext,
+            kind: EscapeSequenceKind,
+            cNode: Character,
+        ) {
+            if (kind === EscapeSequenceKind.hexadecimal) {
+                return
+            }
+
+            const hexEscape = `\\x${cNode.value.toString(16).padStart(2, "0")}`
+
+            context.report({
+                node,
+                loc: getRegexpLocation(cNode),
+                messageId: "expectedHexEscape",
+                data: {
+                    hexEscape,
+                    unexpectedKind: kind,
+                    rejectEscape: cNode.raw,
+                },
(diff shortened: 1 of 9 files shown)


lib/rules/hexadecimal-escape.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        /**
+         * Verify for always
+         */
+        function verifyForAlways(
+            { node, getRegexpLocation, fixReplaceNode }: RegExpContext,
+            kind: EscapeSequenceKind,
+            cNode: Character,
+        ) {
+            if (kind === EscapeSequenceKind.hexadecimal) {
```
RunDevelopment, line 44: I think we should explicitly check for Unicode escapes here. This will also make the octal and control check in `onCharacterEnter` unnecessary.

```suggestion
            if (kind !== EscapeSequenceKind.unicode && kind !== EscapeSequenceKind.unicodeCodePoint) {
```

lib/rules/hexadecimal-escape.ts
```diff
@@ -0,0 +1,125 @@
...
+            context.report({
+                node,
+                loc: getRegexpLocation(cNode),
+                messageId: "unexpectedHexEscape",
+                data: {
+                    hexEscape: cNode.raw,
+                },
+                fix: fixReplaceNode(cNode, () => {
+                    let text = toCharSetSource(cNode.value, flags)
+                    if (isHexadecimalEscape(text)) {
```
RunDevelopment, line 83: `toCharSetSource` might return a character literal. E.g. for `\x41` this will return just `A`.

Since we know that `cNode` is a hex escape, why not always use `` `\\u00${cNode.raw.slice(2)}` ``?

RunDevelopment (changes requested):

lib/rules/hexadecimal-escape.ts
ota-meshi: I was testing `\x0a` and found it strange to be converted to` \u000a`. I personally wanted to convert it to `\n`.
But thanks to your comments, I think this is a personal preference and we may need another rule to convert it to `\n`.
I would like to change this rule to use `\\u00${cNode.raw.slice(2)}`

RunDevelopment: I agree that there should be a rule for converting `\x0a` to `\n`. I'll make an issue.

ota-meshi:
Thank you for your review. I fixed it.

RunDevelopment:
In Unicode patterns, we have the choice between a `\uFFFF` and a `\u{10FFFF}` escape. Should this rule also handle this or should we add another rule for this?

ota-meshi:
What do you think of adding an option to the [prefer-unicode-codepoint-escapes] rule to check more than surrogate pairs?
However, this rule cannot be converted to `\uFFFF`.

Alternatively, I think we could add a new rule for users who like `\uFFFF` and deprecate the [prefer-unicode-codepoint-escapes] rule that conflicts with it rule.

[prefer-unicode-codepoint-escapes]: https://ota-meshi.github.io/eslint-plugin-regexp/rules/prefer-unicode-codepoint-escapes.html

ota-meshi:
I've just realized that conflicts may be avoided if the new rule doesn't check surrogate pairs.

RunDevelopment:
Yes, the new Unicode rule I proposed (let's call it `unicode-escape`for now) doesn't check surrogate pairs. It's supposed to be a sister rule to `hexadecimal-escape`.

When there is a choice between a hex escape and a unicode escape, ``hexadecimal-escape`` forces us to be consistent (e.g. `\x0a` vs `\u000a`). Likewise, when there is a choice between a unicode escape and a unicode codepoint escape, `unicode-escape` forces us to be consistent (e.g. `\u01ff` vs `\u{1ff}`).

ota-meshi:
Sounds good to me! I like the new rule!

RunDevelopment:
Alright, then let's merge this rule add a new one for unicode escapes right after. Do you want to make it or shall I?

RunDevelopment (approved):

ota-meshi:
I want to make it!

PR #164, opened 2021-04-20
Add `regexp/sort-flags` rule

ota-meshi:


lib/rules/sort-flags.ts
```diff
@@ -0,0 +1,181 @@
+import type { Literal } from "estree"
+import type { RegExpVisitor } from "regexpp/visitor"
+import type { RegExpContext } from "../utils"
+import { createRule, defineRegexpVisitor } from "../utils"
+
+export default createRule("sort-flags", {
+    meta: {
+        docs: {
+            description: "require the regex flags to be sorted",
```
RunDevelopment, line 9: I have to correct my English here...

```suggestion
            description: "require regex flags to be sorted",
```

lib/rules/sort-flags.ts
```diff
@@ -0,0 +1,181 @@
...
+            // TODO Switch to recommended in the major version.
+            // recommended: true,
+            recommended: false,
+        },
+        fixable: "code",
+        schema: [
+            {
+                type: "object",
+                properties: {
+                    order: {
```
RunDevelopment, line 19: I think we should not offer this option. For two reasons:

1. We should enforce the order that the [ECMAScript spec](https://tc39.es/ecma262/#sec-get-regexp.prototype.flags) requires for the `RegExp.properties.flags` property.
2. A custom order might interfere with code printers/formatters (like Prettier) that automatically enforce `RegExp.properties.flags` order.

I actually also had this option in the first version of `clean-regex/sort-flags` but removed it for the above reasons.

RunDevelopment (changes requested):

ota-meshi:
Thank you for your review!
I removed the option.

RunDevelopment (approved):

RunDevelopment:
@ota-meshi I have a question: Do we have to worry about non-standard flags? v8 recently added an [`l` flag](https://v8.dev/blog/non-backtracking-regexp). A standard JS parser will refuse to parse regexes with non-standard flags but a parser like Babel might just accept them.

ota-meshi:
@RunDevelopment Thank you for the question!
I don't think the rule needs to handle and check non-standard flags, but I want to avoid the rule crashing due to non-standard flags.

There is a test case to check it.
https://github.com/ota-meshi/eslint-plugin-regexp/blob/master/tests/lib/rules-with-unknown-flag.ts

The typescript-eslint parser accepts any unknown flags.

RunDevelopment:
Thank you for the answer.

I wonder if we should make a rule to disallow non-standard flags. Will such a rule be helpful?

ota-meshi:
Sounds good! I think it helpful for TypeScript users!

RunDevelopment:
Alright, I'll add it :)

PR #166, opened 2021-04-20
Add `regexp/unicode-escape` rule

ota-meshi:
See https://github.com/ota-meshi/eslint-plugin-regexp/pull/163#issuecomment-823177058


RunDevelopment (approved):

RunDevelopment:
Great work! Thank you!

PR #167, opened 2021-04-20
Add `no-non-standard-flag` rule

RunDevelopment:
This implements [this comment](https://github.com/ota-meshi/eslint-plugin-regexp/pull/164#issuecomment-823135188).

Since we now have 3 rules that report/replace flags, I moved the common flags logic into `RegExpContext`.

---

There is still one problem with this PR: the website:

image

We probably have to use `@typescript-eslint/parser` on the website as well. 
@ota-meshi Should we do this? If yes, how?


ota-meshi:
> We probably have to use @typescript-eslint/parser on the website as well.
> @ota-meshi Should we do this? If yes, how?

Hmm.. `@typescript-eslint/parser` requires a very large resource because it requires `typescript`. So I don't want to include `@typescript-eslint/parser` 😅. 
Can you avoid it with example without demo and example with demo using `new RegExp()`?

RunDevelopment:
> demo using `new RegExp()`?

Good idea!

ota-meshi (approved):
LGTM! Thank you!

PR #168, opened 2021-04-20
Add `control-character-escape` rule

### LINKED ISSUE #165: Consistent control escapes
As mentioned by @ota-meshi [here](https://github.com/ota-meshi/eslint-plugin-regexp/pull/163#discussion_r616516223):

> [...] we may need another rule to convert it [`\x0a`] to `\n`.

I want to add that we could make a rule to consistently escape control characters. 

We already have `prefer-t` which does exactly this for `\t`. We should probably deprecate `prefer-t` in favor of a more general rule that handles `\t`, `\r`, `\n`, and `\f`.

(The rule should also be aware of rules like `no-obscure-range`, so we don't have two rules fighting each other.)

RunDevelopment:
This resolves #165.

While `\0` is not a [control escape](https://tc39.es/ecma262/#sec-patterns-static-semantics-character-value), I still added it because it is a control character.

I did not deprecate `prefer-t` here because doing so removes it from the recommended config.

### DIFF
--- a/docs/rules/control-character-escape.md
+++ b/docs/rules/control-character-escape.md
@@ -13,7 +13,7 @@ description: "enforce consistent escaping of control characters"
 
 ## :book: Rule Details
 
-This rule reports ???.
+This rule reports control characters that were not escaped using a control escape (`\0`, `t`, `\n`, `\v`, `f`, `\r`).
 
 <eslint-code-block fix>
 
--- a/docs/rules/control-character-escape.md
+++ b/docs/rules/control-character-escape.md
@@ -1,0 +1,44 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "regexp/control-character-escape"
+description: "enforce consistent escaping of control characters"
+---
+# regexp/control-character-escape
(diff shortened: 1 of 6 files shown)


docs/rules/control-character-escape.md
```diff
@@ -0,0 +1,44 @@
...
+# regexp/control-character-escape
+
+> enforce consistent escaping of control characters
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule reports ???.
```
ota-meshi, line 16: Could you add rule description? Other than that, it looks good!

RunDevelopment: Thank you for pointing that out!

ota-meshi (approved):
Thank you!

PR #171, opened 2021-04-20
Add `prefer-predefined-assertion` rule

RunDevelopment:
This brings over the `clean-regex/prefer-predefines-assertion` rule.


ota-meshi (approved):
LGTM! Thank you!

PR #173, opened 2021-04-21
Maintenance: Update eslintrc and some fix

ota-meshi:


lib/utils/type-tracker/jsdoc/index.ts
```diff
@@ -90,7 +90,7 @@ export class JSDoc {
         }
         const params = (this.params = new JSDocParams())
         for (const param of this.genTags("param")) {
-            const paths: string[] = (param.name || "").split(/\./gu)
+            const paths: string[] = (param.name || "").split(/\./u)
```
RunDevelopment, line 93: Why was the `g` flag removed here?

ota-meshi: `split` seems to search everything without the `g` flag. I knew it for the first time when I looked it up when making rule.

RunDevelopment: Wow. I didn't know that.

RunDevelopment (approved):

PR #174, opened 2021-04-21
Add `regexp/no-useless-lazy` rule that same `regexp/no-useless-non-greedy` rule

ota-meshi:
This PR adds the `regexp/no-useless-lazy` rule that has the same functionality as the `regexp/no-useless-non-greedy` rule.

Unlike #110, this PR does not abolish the `regexp/no-useless-non-greedy` rule. Therefore, this PR can be merged before v1.0.


RunDevelopment:
Were the tests for `no-useless-non-greedy` not removed?

ota-meshi:
Thank you for your review. I accidentally put it back when I rebase.
I will leave only two test cases to check the minimum that the rules can be shared.

RunDevelopment (approved):

PR #175, opened 2021-04-21
Add `regexp/no-standalone-backslash` rule

### LINKED ISSUE #129: New rule to forbid literal backslashes
In non-Unicode mode, the regex `/\c/` is equivalent to `/\\c/`. The `\` in `/\c/` is a literal backslash. It would probably be good to have a rule that detects this rather strange behavior.

ota-meshi:
close #129

### DIFF
--- a/docs/rules/no-standalone-reverse-solidus.md
+++ b/docs/rules/no-standalone-reverse-solidus.md
@@ -14,10 +14,11 @@ description: "disallow standalone reverse solidus (`\\`)"
 
 This rule disallows reverse solidus (`\`) without escape.
 
-The regular expression `/\c/` without the unicode (`u`) flag has the same pattern as `/\\c/`.
+E.g. the regular expression `/\c/` without the unicode (`u`) flag is the same pattern as `/\\c/`.
 
-This is the specification described in [Annex B] of ECMAScript. But very confusing and you should not use it intentionally.
-In most cases you intended the control escape sequence (`\cX`) or another escape, but you maybe made a mistake.
+In most cases, standalone backslashes are used by accident when a control escape sequence (`\cX`) or another escape sequence was intended. They are very confusing and should not be used intentionally.
+
+This behavior is described in [Annex B] of the ECMAScript specification.
 
 [Annex B]: https://tc39.es/ecma262/#sec-regular-expressions-patterns
 
--- a/docs/rules/no-standalone-reverse-solidus.md
+++ b/docs/rules/no-standalone-reverse-solidus.md
@@ -1,0 +1,55 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "regexp/no-standalone-reverse-solidus"
+description: "disallow standalone reverse solidus (`\\`)"
+---
+# regexp/no-standalone-reverse-solidus
+
+> disallow standalone reverse solidus (`\`)
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule disallows reverse solidus (`\`) without escape.
+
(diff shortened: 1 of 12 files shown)


ota-meshi:
There seems to be a problem with GitHub Actions right now. I will re-run it later.
https://www.githubstatus.com/incidents/cj7gzzj30411

lib/rules/no-standalone-reverse-solidus.ts
```diff
@@ -0,0 +1,45 @@
+import type { RegExpVisitor } from "regexpp/visitor"
+import type { RegExpContext } from "../utils"
+import { createRule, defineRegexpVisitor, CP_BACK_SLASH } from "../utils"
+
+export default createRule("no-standalone-reverse-solidus", {
+    meta: {
+        docs: {
+            description: "disallow standalone reverse solidus (`\\`)",
```
RunDevelopment, line 8: This rule should probably be renamed. I had no idea what a "reverse-solidus" was until now. "Backslash" is more commonly used.

docs/rules/no-standalone-reverse-solidus.md
```diff
@@ -0,0 +1,55 @@
...
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule disallows reverse solidus (`\`) without escape.
+
+The regular expression `/\c/` without the unicode (`u`) flag has the same pattern as `/\\c/`.
+
+This is the specification described in [Annex B] of ECMAScript. But very confusing and you should not use it intentionally.
+In most cases you intended the control escape sequence (`\cX`) or another escape, but you maybe made a mistake.
```
RunDevelopment, line 20: I change the text a little.

```suggestion
E.g. the regular expression `/\c/` without the unicode (`u`) flag is the same pattern as `/\\c/`.

In most cases, standalone backslashes are used by accident when a control escape sequence (`\cX`) or another escape sequence was intended. They are very confusing and should not be used intentionally.

This behavior is described in [Annex B] of the ECMAScript specification.
```

RunDevelopment (changes requested):

lib/rules/no-standalone-reverse-solidus.ts
ota-meshi: I personally liked `backslash`, but the name in the spec was `reverse-solidus`. I didn't know which one was easier for people to understand, so I followed the spec.
But you and I think backslash is good so change to `backslash`!

docs/rules/no-standalone-reverse-solidus.md
ota-meshi: Thank you!

ota-meshi:
Thank you for your review! I renamed the rule to `regexp/no-standalone-backslash`.

lib/rules/no-standalone-backslash.ts
```diff
@@ -0,0 +1,45 @@
+import type { RegExpVisitor } from "regexpp/visitor"
+import type { RegExpContext } from "../utils"
+import { createRule, defineRegexpVisitor, CP_BACK_SLASH } from "../utils"
+
+export default createRule("no-standalone-backslash", {
+    meta: {
+        docs: {
+            description: "disallow standalone backslash (`\\`)",
```
RunDevelopment, line 8: It has to be plural here.

```suggestion
            description: "disallow standalone backslashes (`\\`)",
```

RunDevelopment (changes requested):
One last thing and then it's good to go.

lib/rules/no-standalone-backslash.ts
ota-meshi: Thank you!

RunDevelopment (approved):

PR #177, opened 2021-04-22
Updated `isEqual` and `isCovered` by using `regexp-ast-analysis`

RunDevelopment:
I think this is the last rule to refactor for #92, isn't it?


ota-meshi (approved):
Thank you for this PR! That's exactly what I wanted to do!
I merge this PR and close #92.

Issue #178, opened 2021-04-24
`isCoveredNode` false positive


RunDevelopment:
I found a false positive the `isCoveredNode`method produces.

```js
  /(^|\s)[+-]?(?:\d*\.\d+|\d+\.\d*|\d+)(?:e[+-]?\d+)?(?=\s|$)/i
//               ^~~~~~~~ ^~~~~~~~
//                   ^       / 
//                    \_____/ is covered by
```

`isCoveredNode` thinks that `\d+\.\d*` (2nd alt) is covered by `\d*\.\d+` (1st alt).

I found this bug because of a false positive the `no-dupe-disjunction` rule produced because of it.

PR #179, opened 2021-04-24
Updated `no-dupe-disjunctions`

RunDevelopment:
This brings over the functionality of `clean-regex/disjoint-alternatives`.

---

@ota-meshi This PR is still WIP at the moment. I wanted to ask you a few questions (see my comments) before I continue working on this.

---

The results using Prism's 2k regexes. 

The AST approach (just using `isCoveredNode`) found 25 problems:

<details>

```
C:\Users\micha\Git\prism\components\prism-abap.js
  19:4043  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-actionscript.js
  2:316  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-agda.js
  22:76  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-bash.js
   37:23   warning  Unexpected useless alternative. This alternative is already covered by '--?' and can be removed
   regexp/no-dupe-disjunctions
   37:32   warning  Unexpected useless alternative. This alternative is already covered by '\+\+?' and can be removed
   regexp/no-dupe-disjunctions
   37:48   warning  Unexpected useless alternative. This alternative is already covered by '\*\*?' and can be removed
   regexp/no-dupe-disjunctions
   37:87   warning  Unexpected useless alternative. This alternative is already covered by '&&?' and can be removed
   regexp/no-dupe-disjunctions
   37:101  warning  Unexpected useless alternative. This alternative is already covered by '\|\|?' and can be removed
   regexp/no-dupe-disjunctions
  187:36   warning  Unexpected useless alternative. This alternative is already covered by '==?' and can be removed
   regexp/no-dupe-disjunctions
  187:81   warning  Unexpected useless alternative. This alternative is already covered by '\d?[<>]&?' and can be removed  regexp/no-dupe-disjunctions
  187:85   warning  Unexpected useless alternative. This alternative is already covered by '\d?[<>]&?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-concurnas.js
  26:51  warning  Unexpected useless alternative. This alternative is already covered by '\?' and can be removed   regexp/no-dupe-disjunctions
p/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-csharp.js
  274:49  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed  regexp/no-dupe-disjunctions
  311:53  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed  regexp/no-dupe-disjunctions
  314:89  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-dhall.js
  62:43  warning  Unexpected useless alternative. This alternative is already covered by '[!=]=' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-django.js
  33:30  warning  Unexpected useless alternative. This alternative is already covered by '[-+*/%=]=?' and can be removed  regexp/no-dupe-disjunctions
  33:38  warning  Unexpected useless alternative. This alternative is already covered by '[-+*/%=]=?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-factor.js
  65:38  warning  Unexpected useless alternative. This alternative is a subset of '\d*\.\d+' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-lilypond.js
  16:72  warning  Unexpected useless alternative. This alternative is already covered by '[^\s()]' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-parigp.js
  28:113  warning  Unexpected useless alternative. This alternative is already covered by '(?: *<)?(?: *=)?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-psl.js
  22:1361  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-qml.js
  6:35  warning  Unexpected useless alternative. This alternative is already covered by '[^\\()[\]{}"'/]' and can be removed  regexp/no-dupe-disjunctions
  6:55  warning  Unexpected useless alternative. This alternative is already covered by '[^\\()[\]{}"'/]' and can be removed  regexp/no-dupe-disjunctions

✖ 25 problems (0 errors, 25 warnings)
```

</details>

The new NFA approach found 28 problems: (This is an apples-to-apples comparison. The NFA approach actually finds more problems _types_ than the AST approach. These additional problem types are not reported here.)

<details>

```
C:\Users\micha\Git\prism\components\prism-abap.js
  19:4043  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-actionscript.js
  2:316  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-agda.js
  22:76  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-bash.js
   37:23   warning  Unexpected useless alternative. This alternative is already covered by '--?' and can be removed
   regexp/no-dupe-disjunctions
   37:32   warning  Unexpected useless alternative. This alternative is already covered by '\+\+?' and can be removed
   regexp/no-dupe-disjunctions
   37:48   warning  Unexpected useless alternative. This alternative is already covered by '\*\*?' and can be removed
   regexp/no-dupe-disjunctions
   37:87   warning  Unexpected useless alternative. This alternative is already covered by '&&?' and can be removed
   regexp/no-dupe-disjunctions
   37:101  warning  Unexpected useless alternative. This alternative is already covered by '\|\|?' and can be removed
   regexp/no-dupe-disjunctions
  187:36   warning  Unexpected useless alternative. This alternative is already covered by '==?' and can be removed
   regexp/no-dupe-disjunctions
  187:81   warning  Unexpected useless alternative. This alternative is already covered by '\d?[<>]&?' and can be removed  regexp/no-dupe-disjunctions
  187:85   warning  Unexpected useless alternative. This alternative is already covered by '\d?[<>]&?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-cmake.js
  17:1198  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions
  17:2326  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions
  17:3870  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions
  17:4791  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-concurnas.js
  26:51  warning  Unexpected useless alternative. This alternative is already covered by '\?' and can be removed   regexp/no-dupe-disjunctions
  28:30  warning  Unexpected useless alternative. This alternative is already covered by '\w*' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-csharp.js
  274:49  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed
 regexp/no-dupe-disjunctions
  311:53  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed
 regexp/no-dupe-disjunctions
  314:89  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed
 regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-dhall.js
  62:43  warning  Unexpected useless alternative. This alternative is already covered by '[!=]=' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-django.js
  33:30  warning  Unexpected useless alternative. This alternative is already covered by '[-+*/%=]=?' and can be removed  regexp/no-dupe-disjunctions
  33:38  warning  Unexpected useless alternative. This alternative is already covered by '[-+*/%=]=?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-lilypond.js
  16:72  warning  Unexpected useless alternative. This alternative is already covered by '[^\s()]' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-parigp.js
  28:113  warning  Unexpected useless alternative. This alternative is already covered by '(?: *<)?(?: *=)?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-psl.js
  22:1361  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-qml.js
  6:35  warning  Unexpected useless alternative. This alternative is already covered by '[^\\()[\]{}"'/]' and can be removed  regexp/no-dupe-disjunctions
  6:55  warning  Unexpected useless alternative. This alternative is already covered by '[^\\()[\]{}"'/]' and can be removed  regexp/no-dupe-disjunctions

✖ 28 problems (0 errors, 28 warnings)
```

</details>

The hybrid approach found 29 problems:

<details>

```
C:\Users\micha\Git\prism\components\prism-abap.js
  19:4043  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-actionscript.js
  2:316  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-agda.js
  22:76  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-bash.js
   37:23   warning  Unexpected useless alternative. This alternative is already covered by '--?' and can be removed
   regexp/no-dupe-disjunctions
   37:32   warning  Unexpected useless alternative. This alternative is already covered by '\+\+?' and can be removed
   regexp/no-dupe-disjunctions
   37:48   warning  Unexpected useless alternative. This alternative is already covered by '\*\*?' and can be removed
   regexp/no-dupe-disjunctions
   37:87   warning  Unexpected useless alternative. This alternative is already covered by '&&?' and can be removed
   regexp/no-dupe-disjunctions
   37:101  warning  Unexpected useless alternative. This alternative is already covered by '\|\|?' and can be removed
   regexp/no-dupe-disjunctions
  187:36   warning  Unexpected useless alternative. This alternative is already covered by '==?' and can be removed
   regexp/no-dupe-disjunctions
  187:81   warning  Unexpected useless alternative. This alternative is already covered by '\d?[<>]&?' and can be removed  regexp/no-dupe-disjunctions
  187:85   warning  Unexpected useless alternative. This alternative is already covered by '\d?[<>]&?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-cmake.js
  17:1198  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions
  17:2326  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions
  17:3870  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions
  17:4791  warning  Unexpected useless alternative. This alternative is a subset of '\w+_(?:CLANG_TIDY|COMPILER_LAUNCHER|CPPCHECK|CPPLINT|INCLUDE_WHAT_YOU_USE|OUTPUT_NAME|POSTFIX|VISIBILITY_PRESET)' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-concurnas.js
  26:51  warning  Unexpected useless alternative. This alternative is already covered by '\?' and can be removed   regexp/no-dupe-disjunctions
  28:30  warning  Unexpected useless alternative. This alternative is already covered by '\w*' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-csharp.js
  274:49  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed
 regexp/no-dupe-disjunctions
  311:53  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed
 regexp/no-dupe-disjunctions
  314:89  warning  Unexpected useless alternative. This alternative is already covered by '[^"'/()]' and can be removed
 regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-dhall.js
  62:43  warning  Unexpected useless alternative. This alternative is already covered by '[!=]=' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-django.js
  33:30  warning  Unexpected useless alternative. This alternative is already covered by '[-+*/%=]=?' and can be removed
  regexp/no-dupe-disjunctions
  33:38  warning  Unexpected useless alternative. This alternative is already covered by '[-+*/%=]=?' and can be removed
  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-factor.js
  65:38  warning  Unexpected useless alternative. This alternative is a subset of '\d*\.\d+' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-lilypond.js
  16:72  warning  Unexpected useless alternative. This alternative is already covered by '[^\s()]' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-parigp.js
  28:113  warning  Unexpected useless alternative. This alternative is already covered by '(?: *<)?(?: *=)?' and can be removed  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-psl.js
  22:1361  warning  Unexpected duplicate alternative  regexp/no-dupe-disjunctions

C:\Users\micha\Git\prism\components\prism-qml.js
  6:35  warning  Unexpected useless alternative. This alternative is already covered by '[^\\()[\]{}"'/]' and can be removed  regexp/no-dupe-disjunctions
  6:55  warning  Unexpected useless alternative. This alternative is already covered by '[^\\()[\]{}"'/]' and can be removed  regexp/no-dupe-disjunctions

✖ 29 problems (0 errors, 29 warnings)
```

</details>

The interesting bit here is the difference between AST and NFA approach. The one additional problem found by the AST approach is unfortunately a bug (#178). The 4 additional problems found by the NFA approach are true positives and were too difficult to detect for the AST approach.

This should demonstrate how powerful the NFA approach is. However, the NFA approach has its downsides too:

1. It can't handle assertions very well. Most of the time, it can't analyze alternatives with assertions at all. This is why the AST approach is still necessary.
2. Performance. 
    | Approach | Time taken |
    | --- | --- |
    | AST | 5s |
    | NFA | 15s |
    | Hybrid | 20s |

### DIFF
--- a/lib/rules/no-dupe-disjunctions.ts
+++ b/lib/rules/no-dupe-disjunctions.ts
@@ -612,7 +612,7 @@ function deduplicateResults(results: readonly Result[]): Result[] {
     })
 }
 
-enum ReportOption {
+const enum ReportOption {
     all = "all",
     trivial = "trivial",
     interesting = "interesting",
--- a/lib/rules/no-dupe-disjunctions.ts
+++ b/lib/rules/no-dupe-disjunctions.ts
@@ -24,6 +24,7 @@ import {
 } from "refa"
 import type { ReadonlyFlags } from "regexp-ast-analysis"
 import {
+    hasSomeDescendant,
     getMatchingDirection,
     getEffectiveMaximumRepetition,
 } from "regexp-ast-analysis"
@@ -645,15 +646,15 @@ export default createRule("no-dupe-disjunctions", {
         ],
         messages: {
             duplicate:
-                "Unexpected duplicate alternative. This alternative can be removed.{{exp}}",
+                "Unexpected duplicate alternative. This alternative can be removed.{{cap}}{{exp}}",
             subset:
-                "Unexpected useless alternative. This alternative is a strict subset of '{{others}}' and can be removed.{{exp}}",
+                "Unexpected useless alternative. This alternative is a strict subset of '{{others}}' and can be removed.{{cap}}{{exp}}",
             prefixSubset:
-                "Unexpected useless alternative. This alternative is already covered by '{{others}}' and can be removed.",
+                "Unexpected useless alternative. This alternative is already covered by '{{others}}' and can be removed.{{cap}}",
             superset:
-                "Unexpected superset. This alternative is a superset of '{{others}}'. It might be possible to remove the other alternative(s).{{exp}}",
+                "Unexpected superset. This alternative is a superset of '{{others}}'. It might be possible to remove the other alternative(s).{{cap}}{{exp}}",
             overlap:
-                "Unexpected overlap. This alternative overlaps with '{{others}}'. The overlap is '{{expr}}'.{{exp}}",
+                "Unexpected overlap. This alternative overlaps with '{{others}}'. The overlap is '{{expr}}'.{{cap}}{{exp}}",
         },
         type: "suggestion", // "problem",
     },
@@ -680,7 +681,11 @@ export default createRule("no-dupe-disjunctions", {
             const parser = JS.Parser.fromAst({
                 pattern: patternAst,
                 flags: new RegExpParser().parseFlags(
-                    flagsString?.replace(/[^gimsuy]/g, "") ?? "",
+                    [
+                        ...new Set(
+                            (flagsString || "").replace(/[^gimsuy]/g, ""),
+                        ),
+                    ].join(""),
                 ),
             })
 
@@ -698,7 +703,7 @@ export default createRule("no-dupe-disjunctions", {
                     parentNode.alternatives,
                     regexpContext,
                     {
-                        fastAst: false,
+                        fastAst: true,
                         noNfa: false,
                         ignoreOverlap: nodeReport !== ReportOption.all,
                         hasNothingAfter,
@@ -740,6 +745,12 @@ export default createRule("no-dupe-disjunctions", {
                 const exp = isStared(result.alternative)
                     ? " This ambiguity is likely to cause exponential backtracking."
                     : ""
+                const cap = hasSomeDescendant(
+                    result.alternative,
+                    (d) => d.type === "CapturingGroup",
+                )
+                    ? " Careful! This alternative contains capturing groups which might be difficult to remove."
+                    : ""
 
                 const others = result.others.map((a) => a.raw).join("|")
 
@@ -749,7 +760,7 @@ export default createRule("no-dupe-disjunctions", {
                             node,
                             loc: getRegexpLocation(result.alternative),
                             messageId: "duplicate",
-                            data: { exp, others },
+                            data: { exp, cap, others },
                         })
                         break
 
@@ -758,7 +769,7 @@ export default createRule("no-dupe-disjunctions", {
                             node,
                             loc: getRegexpLocation(result.alternative),
                             messageId: "subset",
-                            data: { exp, others },
+                            data: { exp, cap, others },
                         })
                         break
 
@@ -767,7 +778,7 @@ export default createRule("no-dupe-disjunctions", {
                             node,
                             loc: getRegexpLocation(result.alternative),
                             messageId: "prefixSubset",
-                            data: { exp, others },
+                            data: { exp, cap, others },
                         })
                         break
 
@@ -776,7 +787,7 @@ export default createRule("no-dupe-disjunctions", {
                             node,
                             loc: getRegexpLocation(result.alternative),
                             messageId: "superset",
-                            data: { exp, others },
+                            data: { exp, cap, others },
                         })
                         break
 
@@ -787,6 +798,7 @@ export default createRule("no-dupe-disjunctions", {
                             messageId: "overlap",
                             data: {
                                 exp,
+                                cap,
                                 others,
                                 expr: faToSource(result.overlap, flags),
                             },
--- a/lib/rules/no-dupe-disjunctions.ts
+++ b/lib/rules/no-dupe-disjunctions.ts
@@ -136,6 +136,8 @@ function unionAll(nfas: readonly ReadonlyNFA[]): ReadonlyNFA {
     return total
 }
 
+const DFA_OPTIONS: DFA.CreationOptions = { maxNodes: 100_000 }
+
 /**
  * Returns whether one NFA is a subset of another.
  */
@@ -144,8 +146,8 @@ function isSubsetOf(
     subset: ReadonlyNFA,
 ): boolean | null {
     try {
-        const a = DFA.fromIntersection(superset, subset)
-        const b = DFA.fromFA(subset)
+        const a = DFA.fromIntersection(superset, subset, DFA_OPTIONS)
+        const b = DFA.fromFA(subset, DFA_OPTIONS)
         a.minimize()
         b.minimize()
         return a.structurallyEqual(b)
@@ -170,13 +172,13 @@ function getSubsetRelation(
     right: ReadonlyNFA,
 ): SubsetRelation {
     try {
-        const inter = DFA.fromIntersection(left, right)
+        const inter = DFA.fromIntersection(left, right, DFA_OPTIONS)
         inter.minimize()
 
-        const l = DFA.fromFA(left)
+        const l = DFA.fromFA(left, DFA_OPTIONS)
         l.minimize()
 
-        const r = DFA.fromFA(right)
+        const r = DFA.fromFA(right, DFA_OPTIONS)
         r.minimize()
 
         const subset = l.structurallyEqual(inter)
@@ -338,7 +340,7 @@ function* findPrefixDuplicationNfa(
 
     const all = NFA.all({ maxCharacter: alternatives[0][0].maxCharacter })
 
-    for (let i = 1; i < alternatives.length; i++) {
+    for (let i = 0; i < alternatives.length; i++) {
         const [nfa, alternative] = alternatives[i]
 
         const overlapping = alternatives
@@ -446,14 +448,16 @@ function* findDuplication(
     options: Options,
 ): Iterable<Result> {
     // AST-based approach
-    if (options.fastAst) {
-        yield* findDuplicationAstFast(alternatives, context)
-    } else {
-        yield* findDuplicationAst(
-            alternatives,
-            context,
-            options.hasNothingAfter,
-        )
+    if (Math.random() > 6) {
+        if (options.fastAst) {
+            yield* findDuplicationAstFast(alternatives, context)
+        } else {
+            yield* findDuplicationAst(
+                alternatives,
+                context,
+                options.hasNothingAfter,
+            )
+        }
     }
 
     // NFA-based approach
@@ -560,7 +564,7 @@ export default createRule("no-dupe-disjunctions", {
                     {
                         fastAst: false,
                         noNfa: false,
-                        ignoreOverlap: false,
+                        ignoreOverlap: true,
                         hasNothingAfter: hasNothingAfterNode(parentNode),
                         parser,
                     },
@@ -573,6 +577,12 @@ export default createRule("no-dupe-disjunctions", {
 
             /** Report the given result. */
             function report(result: Result) {
+                if (
+                    Math.random() < 4 &&
+                    (result.type === "Overlap" || result.type === "Superset")
+                ) {
+                    return
+                }
                 const exp =
                     getEffectiveMaximumRepetition(result.alternative) > 10
                         ? " This ambiguity is likely to cause exponential backtracking."
--- a/lib/rules/no-dupe-disjunctions.ts
+++ b/lib/rules/no-dupe-disjunctions.ts
@@ -259,10 +259,13 @@ function getSubsetRelation(
  * This function adjusts their subset relation to account for partial NFAs.
  */
 function getPartialSubsetRelation(
-    relation: SubsetRelation,
+    left: ReadonlyNFA,
     leftIsPartial: boolean,
+    right: ReadonlyNFA,
     rightIsPartial: boolean,
 ): SubsetRelation {
+    const relation = getSubsetRelation(left, right)
+
     if (!leftIsPartial && !rightIsPartial) {
         return relation
     }
@@ -497,8 +500,9 @@ function* findDuplicationNfa(
             const others = overlapping.map(([, , a]) => a)
 
             const relation = getPartialSubsetRelation(
-                getSubsetRelation(nfa, othersNfa),
+                nfa,
                 partial,
+                othersNfa,
                 othersPartial,
             )
 
--- a/lib/rules/no-dupe-disjunctions.ts
+++ b/lib/rules/no-dupe-disjunctions.ts
@@ -4,14 +4,16 @@ import type {
     CapturingGroup,
     Group,
     LookaroundAssertion,
+    Node,
     Pattern,
     Quantifier,
 } from "regexpp/ast"
 import type { RegExpContext } from "../utils"
 import { createRule, defineRegexpVisitor } from "../utils"
 import { isCoveredNode, isEqualNodes } from "../utils/regexp-ast"
-import type { Expression, FiniteAutomaton, ReadonlyNFA } from "refa"
+import type { Expression, FiniteAutomaton, NoParent, ReadonlyNFA } from "refa"
 import {
+    TooManyNodesError,
     combineTransformers,
     Transformers,
     DFA,
@@ -28,6 +30,18 @@ import {
 import { RegExpParser } from "regexpp"
 
 type ParentNode = Group | CapturingGroup | Pattern | LookaroundAssertion
+
+/**
+ * Returns whether the node or the elements of the node are effectively
+ * quantified with a star.
+ */
+function isStared(node: Node): boolean {
+    let max = getEffectiveMaximumRepetition(node)
+    if (node.type === "Quantifier") {
+        max *= node.max
+    }
+    return max > 10
+}
 
 /**
  * Check has after pattern
@@ -70,7 +84,7 @@ function hasNothingAfterNode(node: ParentNode): boolean {
 /**
  * Returns whether the given RE AST contains assertions.
  */
-function containsAssertions(expression: Expression): boolean {
+function containsAssertions(expression: NoParent<Expression>): boolean {
     try {
         visitAst(expression, {
             onAssertionEnter() {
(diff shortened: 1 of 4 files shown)


lib/rules/no-dupe-disjunctions.ts
```diff
@@ -9,6 +10,492 @@ import type {
...
+            // Only checking for a subset relation is sufficient here.
+            // Duplicates are VERY unlikely. (Who would use alternatives
+            // like `a|a[^]*`?)
+
+            if (isSubsetOf(othersNfa, nfa)) {
+                yield { type: "PrefixSubset", alternative, others }
+            } else {
+                // The else case is interesting.
+                // It means that _some_ of the paths in the current alternative are useless.
+                // TODO: Decide whether this should be reported as well.
```
RunDevelopment, line 361: What do we want to do here?

In this case, the alternative is partially covered by previous alternatives. Should we report this too? E.g. `\d|\w`, `Foo|\w`.

lib/rules/no-dupe-disjunctions.ts
```diff
@@ -9,6 +10,492 @@ import type {
...
+    right: ReadonlyNFA,
+): SubsetRelation {
+    try {
+        const inter = DFA.fromIntersection(left, right)
+        inter.minimize()
+
+        const l = DFA.fromFA(left)
+        l.minimize()
+
+        const r = DFA.fromFA(right)
```
RunDevelopment, line 179: DFA constructions can blow up exponentially. Refa detects these exponential blowups and throws an error after a certain point. The default for this point is quite conservative at 10k nodes. For perspective, a DFA construction producing 10k nodes takes about 1 to 10ms, and one producing 100k nodes takes about 10 to 100ms, one producing 1M nodes takes about 100 to 1000ms, and so on.

DFA constructions are necessary to figure out which error message to use. If the DFA construction fails, we currently default to the "overlap" error message (instead of reporting e.g. a subset (=useless) alternative). Maybe we should increase the limit and potentially take more time for more accurate error messages?

lib/rules/no-dupe-disjunctions.ts
```diff
@@ -9,6 +10,492 @@ import type {
...
+                    yield { type: "Subset", alternative, others }
+                    break
+
+                case SubsetRelation.leftSupersetOfRight:
+                    yield { type: "Superset", alternative, others }
+                    break
+
+                case SubsetRelation.none:
+                case SubsetRelation.unknown:
+                    // TODO: we might want to report cases where the relation
```
RunDevelopment, line 415: As I said in the [other comment](https://github.com/ota-meshi/eslint-plugin-regexp/pull/179/files#r619676707), the DFA construction might fail. If it fails, we don't know which message to display and currentls default to "overlap".

Should we display a special message if this happens?

The problem might be interesting for the user but I don't want to bother them with implementation details.

RunDevelopment:
@ota-meshi Another question.

The NFA approach also finds supersets and general overlap. While users are probably interested in supersets (e.g. `\b(?:Foo|\w+)\b`), I'm not so sure about overlaps.

An overlap occurs if two alternatives can both accept the same string. E.g. the overlap of `a.*|.*b` is `a.*b`.

Overlaps are _usually_ not harmful and hard to remove (e.g. try to remove the overlap in `a.*|.*b`). However, overlap can cause __exponential backtracking__. 

So my idea was to only report overlaps if they cause exponential backtracking. But this doesn't work because people might use patterns to build other patterns. Simple example:

```js
const foo = /(?:ac?|\wb?)a/ // overlap but not exp. backtracking
const bar = RegExp("^(?:" + foo.source + ")+$") // exp backtracking
```

How should we handle this?

lib/rules/no-dupe-disjunctions.ts
ota-meshi: I think `\d|\w` can delete `\d`, so we may report it. However, I don't think `Foo|\w` can remove one alt. I don't think this rule should be reported. I think it might be nice to have an option to check for partial duplication or another rule.

ota-meshi: Perhaps I misunderstood. Did you say that there is a partial duplication of `(?:\d|\w)|(?:Foo|\w)`?
I think it would be useful to be able to tell the user the correct `\w` location.

lib/rules/no-dupe-disjunctions.ts
ota-meshi: Does this mean there is a possibility of false positives?
I think false negatives are better than false positives, so if we get an error in DFA construction, I think we shouldn't report the error.
I think it is better if you can not limit the number of nodes.

lib/rules/no-dupe-disjunctions.ts
ota-meshi: I don't think DFA construction errors should report a report.

ota-meshi:
I think we shouldn't consider not using `/(?:ac?|\wb?)a/` as a regular expression. 
Not using regular expression literals as regular expressions is almost always a mistake for the user. Perhaps the user should use a string. This is the same as `String(1234)` and I don't think users should use it.

lib/rules/no-dupe-disjunctions.ts
RunDevelopment: > I think it might be nice to have an option to check for partial duplication or another rule.

Then let's add an option. Adding another rule for this would also be possible but we would have to do all the NFA operations again.

> Perhaps I misunderstood.

No, you didn't. :)

lib/rules/no-dupe-disjunctions.ts
RunDevelopment: No, false positives aren't possible.

> I think it is better if you can not limit the number of nodes.

That's not a good idea. DFA construction blows up exponentially, so even a small regex can create a DFA that is gigabytes in size. I don't want your plugin crashing ESLint by eating up all the available memory :)

I'll increase the node limit to 100k.

RunDevelopment:
Firstly, let's give these regex literals that aren't used as regexes a name: let's call them partial patterns or partial regexes.

Over at [Prism](https://github.com/PrismJS/prism), we use partial patterns of the form `/regex/.source`, so this problem is somewhat relevant for us. These partial patterns have a few advantages compared to strings (IDE syntax highlighting, no double escaping for backslashes, early syntax errors) which is why I introduced them to Prism. (Our build process inlines the source, so there's no performance overhead for our users.)

So there are valid uses for partial patterns.

lib/rules/no-dupe-disjunctions.ts
RunDevelopment: Ok. I'll continue to report them as simple overlaps (if allowed by the user).

lib/rules/no-dupe-disjunctions.ts
ota-meshi: > No, false positives aren't possible.

Does that mean you may not be able to report an easy to understand message?
If so, I don't think there is any problem :+1:

RunDevelopment: It means that we might report the wrong _type_ of message. 

I.e. for `A|B`, `B` might be a subset of `A` but if the DFA construction fails, we will report it as "overlap" instead of "subset".

There won't be any cases, where we tell the user that it's safe to remove something when it isn't. But there might be cases where something is safe to remove but we don't tell the user.

ota-meshi:
> (IDE syntax highlighting, no double escaping for backslashes, early syntax errors)

I didn't know the practice. That's an interesting practice.
If you think that should be taken into account, do we want to keep track of variables and not report an error if user only use `.source`? (However, if `.source` is done outside the file, it cannot be tracked.)
I also think users can use the `eslint-disable` comment.

RunDevelopment:
There are some other rules that also assume regex literals to be used as regexes. E.g. I added the `allowTop` option to `no-useless-non-capturing-group` specifically for partial patterns.

Tracking doesn't have to be perfect. In the case of Prism, simply looking for `.source` would be enough. We can always improve it later.

If it's ok with you, then I'll open an issue for this and implement this rule assuming non-partial regexes.

lib/rules/no-dupe-disjunctions.ts
RunDevelopment: I thought about this some more. `\d|\w` is an interesting case and will be reported.

However, `Foo|\w` will not be reported. I mean, the "fixed" pattern would be `Foo|F(?!oo)|[a-zA-EG-Z0-9_]`. There's no point in reporting this.

RunDevelopment:
@ota-meshi Bad news. I have found two bugs in refa that have to be fixed before we can merge this rule. Both bugs only affect DFAs, so all other code using refa is fine. I'm currently working on refa v0.8 and will try to release a new version by the end of the week.

RunDevelopment:
I think this ready for review now. I implemented everything I wanted.

The question now is: What options should this rule have?

I only added two options for now: `report` and `alwaysReportExponentialBacktracking` (see docs). Do we need more? 

I removed the `disallowNeverMatch` option, so this is a breaking change. Should I re-add it in some form until v1.0?

Also, I'm not sure whether it's a good idea to allow `alwaysReportExponentialBacktracking: false`. We are basically allowing our users to hurt themselves.

ota-meshi:
Thank you for lot of work!

> The question now is: What options should this rule have?

I reconsidered it. I think the option should be applied depending on what the user wants to opt out of, but for now I don't think there is any reason to want to opt out, so I don't think we need to offer any more options.
I want we to change to the behavior of `{fastAst: false, noNfa: false,}` that reports the most.

lib/rules/no-dupe-disjunctions.ts
```diff
@@ -22,106 +630,178 @@ export default createRule("no-dupe-disjunctions", {
             {
                 type: "object",
                 properties: {
-                    disallowNeverMatch: { type: "boolean" },
```
ota-meshi, line 25: I want to leave `disallowNeverMatch` in the schema and remove it in v1.0. We don't have to handle `disallowNeverMatch`.

ota-meshi (changes requested):

lib/rules/no-dupe-disjunctions.ts
RunDevelopment: Good idea. Should I log a warning if the user tries to use it?

ota-meshi: I don't think we need log. Users who used or did not use `disallowNeverMatch` will be able to use the enhanced `no-dupe-disjunctions` in the next minor version.
By removing the option from the schema in the v1.0 release, users who use the option in v1.0 will be warned.

lib/rules/no-dupe-disjunctions.ts
```diff
@@ -22,106 +630,180 @@ export default createRule("no-dupe-disjunctions", {
...
+                        ? ReportOption.all
+                        : report
+
+                const hasNothingAfter = hasNothingAfterNode(parentNode)
+
+                const rawResults = findDuplication(
+                    parentNode.alternatives,
+                    regexpContext,
+                    {
+                        fastAst: true,
```
ota-meshi, line 703: Should we change it to `false` to report many warnings?

```suggestion
                        fastAst: false,
```

RunDevelopment: Yes. My bad.

RunDevelopment: I re-enabled the AST approach and added a test case the NFA approach definitely can't detect.

ota-meshi (approved):
I will release a minor version today or tomorrow that includes this change.

PR #188, opened 2021-04-29
Improved `no-assertion-capturing-group`

RunDevelopment:
This brings over half of `clean-regex/no-constant-capturing-group`.

Also, I think we should rename the rule. Even before this update, it also rightfully reported `/()a/` even though `()` obviously doesn't contain assertions. Maybe `no-empty-capturing-group` (because the captured text is always the empty string)?


ota-meshi:
> Even before this update, it also rightfully reported /()a/ even though () obviously doesn't contain assertions.

I didn't intend it in this rule 😅. But I think it's good to rename the rule as you say.

ota-meshi (approved):
LGTM! Thank you!

PR #190, opened 2021-04-30
Added `launch.json`

RunDevelopment:
We already have a `settings.json`, so I added a `launch.json` to make debugging in VSCode easier.


ota-meshi (approved):
Thank you!

PR #197, opened 2021-05-05
Add `no-useless-quantifier` rule

RunDevelopment:
This brings over the `clean-regex/no-unnecessary-quantifier` rule.

---

There is some overlap with `regexp/no-useless-exactly-quantifier`. @ota-meshi, [you said that we might want to deprecate this rule](https://github.com/ota-meshi/eslint-plugin-regexp/issues/91#issuecomment-821790265) and I agree.

However, if we do deprecate `regexp/no-useless-exactly-quantifier`, then we have to handle the zero-quantifier case (e.g. `a{0}`) somewhere else. I think it might be a good idea to add a rule specifically for zero-quantifiers, similar to `clean-regex/no-zero-quantifier`.


ota-meshi (approved):

ota-meshi:
Thank you for this PR!
I would like to add the `no-zero-quantifier` rule and deprecate the `regexp/no-useless-exactly-quantifier` rule:+1:

Issue #181, opened 2021-04-25
Detect partial patterns


RunDevelopment:
Discussion [here](https://github.com/ota-meshi/eslint-plugin-regexp/pull/179#issuecomment-826308528).

RunDevelopment:
I'll review the PR shortly.

> Do you think the `allowTop` option of the `no-useless-non-capturing-group` rule should be changed to ignore only partial patterns? 

I think `allowTop` should have 3 states:

1. `true | "always"`: Always allow top-level groups (partial or not)
2. `false | "never"`: Never allow top-level groups (partial or not)
3. `"partial"` (_default_): Only allow top-level groups in partial patterns.

> Do you think we should check partial patterns for other rules as well?

Yes. `no-lazy-ends` also has a problem with partial patterns. But I think those 3 rules are the only ones that have to account for partial patterns.

ota-meshi:
I opened two PRs. #215, #216

If we can merge both, I think we can close this issue.

RunDevelopment:
Thank you @ota-meshi!

RunDevelopment:
Thanks to your awesome work @ota-meshi, we can now close this issue.

PR #204, opened 2021-05-11
Add options related to reporting partial patterns to `regexp/no-dupe-disjunctions` rule

ota-meshi:
This PR changes the `regexp/no-dupe-disjunctions` rule so that it does not report `foo` regex in the next case.

```js
  const foo = /(?:ac?|\wb?)a/ // overlap but not exp. backtracking
  const bar = RegExp("^(?:"+foo.source+")+$") // exp backtracking
```

Also, each rule in the plugin now verifies the `bar` regex. Because it now tracks RegExp#source.

---

related #181.


ota-meshi:
Thank you for your review! I changed it. Could you check it again?

lib/utils/is-partial-pattern.ts
```diff
@@ -0,0 +1,224 @@
...
+} from "estree"
+import { findVariable, getParent, getStringIfConstant } from "./ast-utils"
+
+/**
+ * Check whether the given expression node is using it only as a partial pattern (uses only `.source`).
+ */
+export function isPartialPattern(
+    node: Expression,
+    context: Rule.RuleContext,
+): boolean {
```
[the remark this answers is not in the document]
ota-meshi: `unknown` is added because there are cases where the usage of the pattern is unknown or it is not used at all.

lib/rules/no-dupe-disjunctions.ts
```diff
@@ -789,17 +791,22 @@ export default createRule("no-dupe-disjunctions", {
...
-                            data: {
-                                exp,
-                                cap,
-                                others,
-                                expr: faToSource(result.overlap, flags),
-                            },
-                        })
+                        if (
+                            isStared(result.alternative) ||
+                            getUsageOfPattern() !== UsageOfPattern.partial
```
RunDevelopment, line 796: This is the wrong place to check for partial patterns. The `reportResult` has the sole purpose of reporting.

Also, this is wrong. We detect partial patterns to report overlaps that might cause exp backtracking. Example:

```ts
const foo = /[ab]|[bc]/.source;
const bar = RegExp(`(${foo})+$`);
```

However, I'm not sure what the _correct_ behavior would be. The only thing I know for certain is:

- `report: "all"` should not be affected by partial patterns. `report: "all"` should just report everything it finds.

But there are still unanswered questions:

- Partial patterns and `PrefixSubset` results. E.g. ``var a = /a|aa/.source; var b = RegExp(`\\b${a}\\b`);``. In partial patterns, `PrefixSubset` results _might_ be false positives. We can't know for certain. Should we report them?

- _Potential_ cases of exp backtracking. We don't know whether overlap in partial patterns causes exponential backtracking. It can (see example above) but it also might not. We can't be certain.

  One solution might be to add another option like `reportPotentialExponentialBacktracking`. If set to `true`, all potential cases of exponential backtracking in partial patterns will be reported.

  Another solution is to remove the `alwaysReportExpoentialBacktracking: boolean` option (let's call this option `AREB` from now on...) and replace it with a new option `reportExpoentialBacktracking: "none" | "certain" | "potential"`. `none` is the current behavior of `AREB: false`, `certain` is the current behavior of `AREB: true`, and `potential` is the `AREB: true` + the potential cases of partial patterns.

Thoughts @ota-meshi?

RunDevelopment (changes requested):
I like the new `getUsageOfPattern` method. Nice work!

However, we have to talk about how the `no-dupe-disjunctions` rule uses the new pattern usage information.

lib/rules/no-dupe-disjunctions.ts
ota-meshi: I like the idea of `reportExpoentialBacktracking`. I will implement the `reportExpoentialBacktracking` option.

> Partial patterns and PrefixSubset results. E.g. `var a = /a|aa/.source; var b = RegExp(`\\b${a}\\b`);`. In partial patterns, PrefixSubset results might be false positives. We can't know for certain. Should we report them?

I didn't notice it. We probably need a similar option, `reportPrefixSubset: "certain" | "potential"`.

I will fix a bug that doesn't check template literals.

ota-meshi:
> report: "all" should not be affected by partial patterns. report: "all" should just report everything it finds.

I missed it. I will fix it again.

ota-meshi:
I changed this PR. Would you please check it again?

docs/rules/no-dupe-disjunctions.md
```diff
@@ -87,11 +88,78 @@ This option control what types of duplications will be reported. The possible va
...
   If your codebase contained many such partial regexes, then reporting all cases might yield cases that could not be identified as causing exponential backtracking.
 
-### `alwaysReportExponentialBacktracking: boolean`
+### `reportExponentialBacktracking`
 
-If set to `true`, then this rule will always report partial duplications that can cause exponential backtracking. This option is set to `true` by default.
+Partial duplications (overlaps) are only reported by `report: "all"` even though they sometimes cause exponential backtracking. This option will force the other `report` modes to also report partial duplications if the partial duplications are likely to cause exponential backtracking.
 
-Only set this option to `false` if you have some other mean to reliably detect exponential backtracking.
+- `reportExponentialBacktracking: "potential"` (_default_)
```
RunDevelopment, line 95: Regarding the defaults for `reportExponentialBacktracking` and `reportUnreachable`:

Right now both are set to `"potential"` which will produce some false positives. I worry that users might get annoyed by the over-reporting and turn off the rule.

I think we should change the default for `reportUnreachable` to `"certain"`. It's not catastrophically bad to have some false negatives here.

As for `reportExponentialBacktracking`, I think `potential` is okay. Missing cases can be catastrophic, so I think it's okay to have some false positives.

ota-meshi: I agree that `reportUnreachable: "certain"` should be the default. I will change it.

RunDevelopment (approved):

PR #206, opened 2021-05-12
Added rule categories

### LINKED ISSUE #198: Categorize rules
The list of supported rules is getting long (56 rules) and will only get longer in the future. 

I think we should categorize our rules so our users can quickly find rules with somewhat similar functionality.

For now, I think the categories "Possible Errors", "Best Practices", and "Stylistic Issues" (taken from the ESLint website) might be a good fit.

@ota-meshi What do you think about this?

RunDevelopment:
This adds rule categories as discussed in #198.

The website now looks like this:

image

---

This resolves #198


RunDevelopment:
I just went and assigned categories based on what I thought felt right. Feel free to suggest changes @ota-meshi.

ota-meshi (approved):
Thank you for this PR! I like it categories.

PR #207, opened 2021-05-12
Add `optimal-quantifier-concatenation` rule

RunDevelopment:
This brings over `clean-regex/optimal-concatenation-quantifier`.

If renamed the `clean-regex` rule and extended its functionality a little. While the old `clean-regex` rule was able to fix `aa*` -> `a+`, the new rule is also able to fix `a{4}a+` -> `a{5,}`.

As you can see from the examples above, there is some overlap with the `prefer-quantifier` rule. To resolve this, I think we should simplify the `prefer-quantifier` to only handle character cases and ignore quantifiers. The `prefer-quantifier` rule will then _create_ quantifiers but it won't _modify_ existing ones. I will do this change in another PR.


ota-meshi (approved):
LGTM! Thank you!

PR #208, opened 2021-05-13
Simplify `prefer-quantifier`

RunDevelopment:
This PR changes `prefer-quantifier` so its functionality no longer overlaps with `optimal-quantifier-concatenation`.

This is the PR I promised in #207.


ota-meshi (approved):
LGTM!

PR #210, opened 2021-05-18
#204 changes

RunDevelopment:
I made a few changes to #204.

Changes:

- Added more documentation. 

- Renamed `reportPrefixSubset` -> `reportUnreachable`. The reported alternatives are unreachable so let's call it that. The "prefix suffix" name is also correct but it's a lot more technical and difficult to understand for our users.

- `report: "all"` and `reportUnreachable: "certain"`: Your initial implementation before my comment "`report: "all"` should just report everything it finds" was better. My comment was wrong and I correct this here.

- I moved the logic that processes options and decides which cases are reported into two functions: `getFilterInfo` and `filterResults`. This separates how filtering is done from determining static properties of the analyses node (e.g. whether the node is under a star quantifier).

### DIFF
--- a/docs/rules/no-dupe-disjunctions.md
+++ b/docs/rules/no-dupe-disjunctions.md
@@ -45,7 +45,7 @@ var foo = /.|a|b|c/
     {
         "report": "trivial",
         "reportExponentialBacktracking": "potential",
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
Thank you!

PR #211, opened 2021-05-19
#204 dedupe fix

RunDevelopment:
The result deduplication is an issue now. Example:

```js
var partial = /a?|a+/.source; // overlap between `a?` and `a+` is `a`
var whole = RegExp(`^(?:${partial})+$`);
```

Let's assume that all potential cases of exp backtracking and unreachable alternatives are configured to be reported. What actually gets reported here?

We will report the unreachable alternative `a+` for `partial` and exponential backtracking because of the overlap in `whole`. But we don't report the potential exp backtracking in `partial`! This is a problem because `whole` might not be in the same file as `partial`, so we might not be able to detect the exp backtracking in `whole`.

Even worse, not only do we not warn the user about the potential exp backtracking in `partial`, we even report a false positive instead. The user might even turn off this rule for the `partial` regex because of the false positive...

This problem only happens if a `PrefixSubset` result overshadows another result for the same alternative. The fix is to report one addition `Superset` or `Overlap` if the first reported result of a `PrefixSubset`.

### DIFF
--- a/lib/rules/no-dupe-disjunctions.ts
+++ b/lib/rules/no-dupe-disjunctions.ts
@@ -590,26 +590,39 @@ const RESULT_TYPE_ORDER: Result["type"][] = [
  * Returns an array of the given results that is sorted by result type from
  * most important to least important.
  */
-function sortResultTypes(unsorted: Iterable<Result>): Result[] {
-    return [...unsorted].sort(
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
Thank you!

PR #215, opened 2021-05-28
Add option to allow only partial use to the `allowTop` option of `regexp/no-useless-non-capturing-group` rule.

ota-meshi:
Related to https://github.com/ota-meshi/eslint-plugin-regexp/issues/181#issuecomment-838271791.

### DIFF
--- a/lib/rules/no-useless-non-capturing-group.ts
+++ b/lib/rules/no-useless-non-capturing-group.ts
@@ -72,27 +72,24 @@ export default createRule("no-useless-non-capturing-group", {
             fixReplaceNode,
             getUsageOfPattern,
         }: RegExpContext): RegExpVisitor.Handlers {
-            let isIgnore: (gNode: Group) => boolean
+            let isIgnored: (gNode: Group) => boolean
             if (allowTop === "always") {
-                isIgnore = isTopLevel
+                isIgnored = isTopLevel
             } else if (allowTop === "partial") {
                 const usageOfPattern = getUsageOfPattern()
-                if (
-                    usageOfPattern === UsageOfPattern.partial ||
-                    usageOfPattern === UsageOfPattern.mixed
-                ) {
-                    isIgnore = isTopLevel
+                if (usageOfPattern !== UsageOfPattern.whole) {
+                    isIgnored = isTopLevel
                 } else {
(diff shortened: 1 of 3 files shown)


lib/rules/no-useless-non-capturing-group.ts
```diff
@@ -56,10 +70,29 @@ export default createRule("no-useless-non-capturing-group", {
             node,
             getRegexpLocation,
             fixReplaceNode,
+            getUsageOfPattern,
         }: RegExpContext): RegExpVisitor.Handlers {
+            let isIgnore: (gNode: Group) => boolean
```
RunDevelopment, line 75: Pedantic nit...

```suggestion
            let isIgnored: (gNode: Group) => boolean
```

lib/rules/no-useless-non-capturing-group.ts
```diff
@@ -56,10 +70,29 @@ export default createRule("no-useless-non-capturing-group", {
...
+            if (allowTop === "always") {
+                isIgnore = isTopLevel
+            } else if (allowTop === "partial") {
+                const usageOfPattern = getUsageOfPattern()
+                if (
+                    usageOfPattern === UsageOfPattern.partial ||
+                    usageOfPattern === UsageOfPattern.mixed
+                ) {
+                    isIgnore = isTopLevel
```
RunDevelopment, line 84: I would prefer this to be more conservative. The usage might be `UsageOfPattern.unknown`. If we don't know the usage, then we shouldn't assume that it's okay to remove the top-level group.

```suggestion
                if (usageOfPattern !== UsageOfPattern.whole) {
                    isIgnore = isTopLevel
```

RunDevelopment (changes requested):
Thank you @ota-meshi!

Two minor things but apart from that good to go.

ota-meshi:
Thank you for your review! I updated it, so could you check it again?

ota-meshi:
I noticed that I haven't changed about autofix yet. I will fix it again.

ota-meshi:
>I noticed that I haven't changed about autofix yet. I will fix it again.

I fixed it.

RunDevelopment:
Thank you @ota-meshi!

PR #216, opened 2021-05-29
Add `ignorePartial` option to `regexp/no-lazy-ends` rule

ota-meshi:
This PR adds an `ignorePartial` option to the `regexp/no-lazy-ends` rule.

If user specify `true` for the `ignorePartial` option, the rule ignores regular expressions that are partially used as fragments.

Related to #181.

### DIFF
--- a/lib/rules/no-lazy-ends.ts
+++ b/lib/rules/no-lazy-ends.ts
@@ -2,6 +2,7 @@ import type { RegExpVisitor } from "regexpp/visitor"
 import type { Alternative, Quantifier } from "regexpp/ast"
 import type { RegExpContext } from "../utils"
 import { createRule, defineRegexpVisitor } from "../utils"
+import { UsageOfPattern } from "../utils/get-usage-of-pattern"
 
 /**
  * Extract lazy end quantifiers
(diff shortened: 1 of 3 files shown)


lib/rules/no-lazy-ends.ts
```diff
@@ -64,13 +73,27 @@ export default createRule("no-lazy-ends", {
...
         function createVisitor({
             node,
             getRegexpLocation,
+            getUsageOfPattern,
         }: RegExpContext): RegExpVisitor.Handlers {
+            if (ignorePartial) {
+                const usageOfPattern = getUsageOfPattern()
+                if (
+                    usageOfPattern === UsageOfPattern.partial ||
+                    usageOfPattern === UsageOfPattern.mixed
```
RunDevelopment, line 90: Should we [also make this conservative](https://github.com/ota-meshi/eslint-plugin-regexp/pull/215#discussion_r642066441)?

I think that `no-useless-non-capturing-group` should be conservative because it's auto-fixable, so false positives can cause harm directly. But `no-lazy-ends` isn't auto-fixable. The user has to decide each case. Maybe false positives are more acceptable here?

ota-meshi:
Thank you for your review! I updated it, so could you check it again?

RunDevelopment (approved):

RunDevelopment:
Thank you @ota-meshi!

PR #219, opened 2021-06-01
Fixed `prefer-range` for adjacent bot not-allowed ranges

### LINKED ISSUE #218: `prefer-range` combines adjacent but not-allowed ranges.
With the settings:

```js
settings: {
	regexp: {
		// allow alphanumeric and cyrillic ranges
		allowedCharacterRanges: ['alphanumeric', 'а-я', 'А-Я']
	}
},
```

`prefer-range` will combine `[а-яА-Я]` into `[А-я]` which then gets reported by `no-obscure-range` as an error. Clearly, `perfer-range` is a little too eager.

RunDevelopment:
This fixes #218.

### DIFF
--- a/lib/rules/prefer-range.ts
+++ b/lib/rules/prefer-range.ts
@@ -103,11 +103,25 @@ export default createRule("prefer-range", {
                         } else {
                             continue
                         }
-                        const group = groups.find(
-                            (gp) =>
+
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
LGTM! Thank you!

Issue #194, opened 2021-05-01
Rename rule no-assertion-capturing-group -> no-empty-capturing-group


ota-meshi:
See #188 comments.

PR #221, opened 2021-06-02
Add `regexp/no-empty-capturing-group` rule that same `regexp/no-assertion-capturing-group` rule 

ota-meshi:
This PR adds the `regexp/no-empty-capturing-group` rule that has the same functionality as the `regexp/no-assertion-capturing-group` rule.

related to #194.


RunDevelopment (approved):

RunDevelopment:
Sorry for the late review!

PR #224, opened 2021-06-02
Add support for `y` flag to `regexp/no-useless-flag` rule.

### LINKED ISSUE #202: Useless `y` flag
I just learned that [`RegExp.prototype [ @@split ]`](https://tc39.es/ecma262/#sec-regexp.prototype-@@split) automatically adds a `y` flag to the regexp. As noted:

> The `@@split` method ignores the value of the "global" and "sticky" properties of this RegExp object.

This means that `"str".split(/re/)` and `"str".split(/re/gy)` are equivalent. The `no-useless-flag` rule current does report the `g` flag but not the `y` flag.

ota-meshi:
close #202

### DIFF
--- a/tests/lib/rules/no-useless-flag.ts
+++ b/tests/lib/rules/no-useless-flag.ts
@@ -106,6 +106,80 @@ tester.run("no-useless-flag", rule as any, {
         unknown.exec(regex)
         `,
 
+        // y
+        `
+        const regex = /foo/y
+        regex.lastIndex = 4
+        regex.test('bar_foo')
+        `,
+        `
+        const regex = /foo/y;
+        regex.test(bar);
+        regex.test(bar);
+        `,
+        `
+        const regex = /foo/y;
+        regex.exec(bar);
+        regex.exec(bar);
+        `,
+        `
+        const regex = /foo/y;
+        `,
+        `
+        const regex = /foo/y;
+        function fn () {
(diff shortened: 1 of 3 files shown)


tests/lib/rules/no-useless-flag.ts
```diff
@@ -393,6 +467,85 @@ tester.run("no-useless-flag", rule as any, {
...
+                },
+            ],
+        },
+        {
+            code: `
+            const a = /foo/y;
+            const b = a;
+            const regex = b;
+
+            'str'.split(b)
```
RunDevelopment, line 537: Is this really correct?

I thought the default source mode is "script" which means that `a`, `b`, and `regex` are all global variables and might be used in another script where the `y` might be relevant.

ota-meshi: I searched for it. ESLint seems to judge the variable declared by the exported directive as a global variable.

https://eslint.org/docs/rules/no-unused-vars#exported

We should support `exported` too, but the problem is not limited to `no-useless-flag`. If `getUsageOfPattern` is also assigned to the exported variable, it should return `unknown`.
I would like to work on this issue as a separate issue.

RunDevelopment (approved):

RunDevelopment:
Thank you @ota-meshi!

PR #220, opened 2021-06-02
Add `regexp/strict` rule

ota-meshi:
Related to https://github.com/ota-meshi/eslint-plugin-regexp/issues/91#issuecomment-852221177.

PR #225, opened 2021-06-02
Improved `regexp/strict` rule

RunDevelopment:
This adds the logic I was talking about [here](https://github.com/ota-meshi/eslint-plugin-regexp/pull/220#issuecomment-852858869). Many reports are fixable. Only reports that are likely a but are not fixable.

However, there is still one problem with this: the tests don't pass because `RegExpValidator` says that `/(?<foo>)\k<foo>/` isn't valid. I have no idea why it rejects that regex. Do you know why @ota-meshi?

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ The rules with the following star :star: are included in the `plugin:regexp/reco
 | [regexp/no-useless-assertions](https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-useless-assertions.html) | disallow assertions that are known to always accept (or reject) |  |
 | [regexp/no-useless-backreference](https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-useless-backreference.html) | disallow useless backreferences in regular expressions | :star: |
 | [regexp/no-useless-dollar-replacements](https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-useless-dollar-replacements.html) | disallow useless `$` replacements in replacement string |  |
-| [regexp/strict](https://ota-meshi.github.io/eslint-plugin-regexp/rules/strict.html) | disallow not strictly valid regular expressions |  |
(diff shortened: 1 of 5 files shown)


ota-meshi:
> `/(?<foo>)\k<foo>/` isn't valid

I think it's probably an invalid regular expression because of backward compatibility.

> However, the named backreference syntax, `/\k<foo>/`, is currently permitted in non-Unicode RegExps and matches the literal string `"k<foo>"`. In Unicode RegExps, such escapes are banned.

https://github.com/tc39/proposal-regexp-named-groups#backwards-compatibility-of-new-syntax

I'm not sure if it's valid as a strict non-Unicode regular expression.

RunDevelopment:
v8 will happily execute that regex. Example:

```
Node.js v14.15.4.
> /^(?<foo>A)\k<foo>$/.test("AA")
true
```

It _should_ be valid because:

> In this proposal, `\k<foo>` in non-Unicode RegExps will continue to match the literal string `"k<foo>"` _unless_ the RegExp contains a named group, in which case it will match that group or be a syntax error [...]

And I think, I figured out why regexpp throws an error. It's a bug in regexpp's validator: mysticatea/regexpp#23.

Now we have two option:

1) Wait until the bug is fixed.
2) Publish `regexp/strict` without `RegExpValidator` and re-add it later when the bug is fixed.

Which one do we want to choos?

ota-meshi:
I heard a rumor that mysticatea seems to be busy these days and doesn't have time to maintain OSS.
I don't know if he will reply, but I will try to send him a message via chat.

RunDevelopment:
I see. Thank you @ota-meshi!

In that case, I'll disable the `RegExpValidator`-based check for patterns with named backreferences. We can re-add it as soon as the bug is fixed.

ota-meshi (approved):
Thank you very much!

PR #226, opened 2021-06-02
Make `no-useless-escape` fixable

RunDevelopment:
I made the `no-useless-escape` rule fixable. All [stylistic rules](https://ota-meshi.github.io/eslint-plugin-regexp/rules/#stylistic-issues) are now fixable.

Reports for potential escape sequences (e.g. `\x;`) will not be fixed. Invalid escape sequences are a potential error and not a stylistic issue, so they shouldn't be auto-fixable.

### DIFF
--- a/lib/rules/no-useless-escape.ts
+++ b/lib/rules/no-useless-escape.ts
@@ -46,7 +46,7 @@ const REGEX_ESCAPES = new Set([
     CP_CLOSING_PAREN, // )
 ])
 
-const POTENTIAL_ESCAPE_SEQUENCE = new Set("uxkpP".split(""))
+const POTENTIAL_ESCAPE_SEQUENCE = new Set("uxkpP")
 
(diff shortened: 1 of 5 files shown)


lib/rules/no-useless-escape.ts
```diff
@@ -46,6 +46,8 @@ const REGEX_ESCAPES = new Set([
     CP_CLOSING_PAREN, // )
 ])
 
+const POTENTIAL_ESCAPE_SEQUENCE = new Set("uxkpP".split(""))
```
ota-meshi, line 49: It is the same without `.split("")`.

```suggestion
const POTENTIAL_ESCAPE_SEQUENCE = new Set("uxkpP")
```

ota-meshi (approved):

ota-meshi (approved):
Thank you!

PR #234, opened 2021-06-08
Change to handle variables listed in `exported` directive as unknown usage

### LINKED ISSUE #227: Support `exported` directive
We track variables with some rules to determine how to use regular expressions, but we have to decide that variables declared by the `exported` directive will be used externally and handle them as unclear.

https://eslint.org/docs/rules/no-unused-vars#exported

https://github.com/ota-meshi/eslint-plugin-regexp/pull/224#discussion_r644426789

ota-meshi:
- The `no-unused-capturing-group` rule, the `no-useless-flag` rule, and the rules using `getUsageOfPattern`, have been changed to handle the `exported` directive.
- Variables were tracked by the `no-unused-capturing-group` rule, the `no-useless-flag` rule, and the `getUsageOfPattern` function. In this PR, variable tracking is shared with the `extractExpressionReferences` function, and it has been changed to use this function.
- The `no-unused-capturing-group` rule has improved iteration tracking.
- The tracking of used properties that the `no-unused-capturing-group` rule was handling has been separated by the `extractPropertyReferences` function. This can also be used to detect the `d` flag, which is probably unused.

close #227


lib/utils/ast-utils/extract-expression-references.ts
```diff
@@ -0,0 +1,312 @@
...
+    functions: Map<
+        FunctionDeclaration | FunctionExpression | ArrowFunctionExpression,
+        Set<number>
+    >
+}
+/** Extract references from the given expression */
+export function* extractExpressionReferences(
+    node: Expression,
+    context: Rule.RuleContext,
+): IterableIterator<ExpressionReference> {
```
RunDevelopment, line 75:
```suggestion
): Iterable<ExpressionReference> {
```

A little suggestion here. It's correct to return an `IterableIterator` here but this might make it harder to change the function in the future. If we ever re-implement `extractExpressionReferences` as a normal function, it will be hard to guarantee the `IterableIterator` interface. Plus, only the `Iterable` part of the `IterableIterator` interface is typically used.

I avoid the `IterableIterator` in my project for those reasons. If I really need an iterator, I can always just all `iterable[Symbol.iterator]()` anyway.

ota-meshi: Thank you for your review!
I hadn't really thought about the difference between them. I will change this PR.

RunDevelopment:
Good work @ota-meshi! Apart from my above comment, I don't have much to say about this PR. Merge this anytime you think it's ready.

RunDevelopment (changes requested):
Two nits

ota-meshi:
Thank you for checking! I changed this PR.

RunDevelopment (approved):

PR #235, opened 2021-06-08
Add `sort-alternatives` rule

### LINKED ISSUE #214: Sort word lists rule
Some regexes contain (potentially long) lists of words. E.g. almost every language over at Prism has at least one regex [like this](https://github.com/PrismJS/prism/blob/01cc18eda9e59198ab2f596944fc3338ac85a2c0/components/prism-aql.js#L18).

For many of those lists, the order of alternatives doesn't matter. I.e. most word lists in Prism are surrounded with `\b`s. In this case, we are free to reorder (=sort) the alternatives.

Sorting these lists is beneficial because it makes it easier for humans to find words. This makes it easier to maintain the regexes containing these lists.

Another additional advantage is that compression algorithms (e.g. gzip) are typically able to better compress sorted lists. I tested this with a few examples and sorted lists were around 10% smaller when compressed with gzip.

---

This rule is basically [`order-in-character-class`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/order-in-character-class.html) but for groups.

RunDevelopment:
This resolves #214.

This adds the `sort-alternatives` rule. 

I constrained the rule to only sort alternatives containing certain characters and non-star quantifiers. We might want to change these constraints in the future but they seem to be a good balance right now.

The implementation will be able to handle any change in constraints. Even without any constraints, it will still work correctly (= it won't sort alternatives if it changes the pattern) but the sorting might not make much sense (e.g. it will sort lists of punctuation characters). 

One important note: numbers get special treatments. List (and sublists) containing only numbers will be sorted by value and not lexicographically. E.g. `3|5|16|1|32` will be sorted as `1|3|5|16|32` and not `1|16|3|32|5`.

### DIFF
--- a/lib/rules/sort-alternatives.ts
+++ b/lib/rules/sort-alternatives.ts
@@ -1,8 +1,17 @@
 import type { RegExpVisitor } from "regexpp/visitor"
-import type { Alternative, CapturingGroup, Element, Group } from "regexpp/ast"
+import type {
+    Alternative,
+    CapturingGroup,
+    Element,
+    Group,
+    LookaroundAssertion,
+    Pattern,
+} from "regexpp/ast"
 import type { RegExpContext } from "../utils"
-import { createRule, defineRegexpVisitor } from "../utils"
+import { CP_MINUS, CP_SPACE, createRule, defineRegexpVisitor } from "../utils"
+import type { ReadonlyFlags } from "regexp-ast-analysis"
 import {
+    getLengthRange,
     Chars,
     getFirstCharAfter,
     getFirstConsumedChar,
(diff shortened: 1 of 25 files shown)


lib/rules/sort-alternatives.ts
```diff
@@ -0,0 +1,396 @@
...
+
+    alternatives.sort((a, b) => {
+        const firstA = firstChars.get(a)!
+        const firstB = firstChars.get(b)!
+        if (firstA !== firstB) {
+            return firstA - firstB
+        }
+
+        if (context.flags.ignoreCase) {
+            return a.raw.toUpperCase().localeCompare(b.raw.toUpperCase())
```
ota-meshi, line 173: What do you think about making the following change?

```suggestion
            return (
                a.raw.toUpperCase().localeCompare(b.raw.toUpperCase()) ||
                a.raw.localeCompare(b.raw)
            )
```

With this change, the following two patterns enforce a consistent order:

`/\b(?:A|a|C|c|B|b)\b/i` -> `/\b(?:a|A|b|B|c|C)\b/i`
`/\b(?:a|A|C|c|b|B)\b/i` -> `/\b(?:a|A|b|B|c|C)\b/i`

Before the change, it returned the following results:

`/\b(?:A|a|C|c|B|b)\b/i` -> `/\b(?:A|a|B|b|C|c)\b/i`
`/\b(?:a|A|C|c|b|B)\b/i` -> `/\b(?:a|A|b|B|C|c)\b/i`

However, after the change, the following pattern will result in an error. This may be annoying to the user.

`/\b(?:A|a|B|b|C|c)\b/i`

RunDevelopment: Very good point. You also just pointed out a bug. 

Right now, alternatives are sorted by the first character first and then by raw. But the first character is ordered by byte order (by ascending Unicode code point value; "A" < "a"), and the raw is sorted by the current locale ("a" < "A").

I will use your suggestion and implement a byte-order compare function.

ota-meshi (approved):
LGTM!

Issue #159, opened 2021-04-19
Detect super-linear worst-case runtime


RunDevelopment:
I think we should add a rule that reports cases for exponential backtracking and polynomial backtracking.

However, there are two problems standing in our way:

### Performance

Detecting exponential backtracking is computationally expensive. 

Static analysis methods might only take a few milliseconds per regex but this will quickly add up for a large codebase (e.g. Prism has about 2k regexes). Fuzzers will usually take about one second or more per regex and are completely impractical for our purposes.

#### Solutions

- Since the detection itself is deterministic, a simple cache might largely solve the performance problem.

    However, I am not aware of ESLint providing such a feature.

### Detection

I am not aware of any existing detector implementation in JavaScript (time of writing: 2020-04-19).

[`safe-regex`](https://www.npmjs.com/package/safe-regex) isn't practical. It implements a detection method that simply doesn't work. Star height has little to do with exponential runtime. I.e. `(a|a)*b` has a star height of 1 and the `(a|a)*` will cause exponential backtracking, and `(ab+)*c` has a star height of 2 without any form of exponential backtracking.

#### Solutions

- Right now, we could use my library [scslre](https://github.com/RunDevelopment/scslre) to detect at least some polynomial and exponential backtracking quickly. scslre even offers fixes for simple cases.

- There is also [this method](https://github.com/PrismJS/prism/blob/80d475e4d8ddc8c5d744a4143b3078545d72d6aa/tests/pattern-tests.js#L436-L530) I implemented for Prism. This method is currently used in PrismJS and Highlight.JS. The upside of this method is that it is quick. It can analyze the >2.5k regexes of Prism in less than 5 seconds. The downside of this method is that there are false negatives and it cannot offer fixes. However, the positives it does find are mostly true positives.

    I should note that half of this method is already partially implemented by the `no-dupe-disjunctions` rule. If we find two alternatives that are not disjoint and both are children of a star quantifier, then we can reasonably assume that this will cause exponential backtracking (e.g. `(a|a)*`). Changing the `no-dupe-disjunctions` to add this information should be fairly easy.

- Given that [refa](https://github.com/RunDevelopment/refa) recently added [ENFA](https://rundevelopment.github.io/refa/docs/dev/classes/enfa.html)s (not published yet), someone could probably implement a detection method like [RXXR](https://www.cs.bham.ac.uk/~hxt/research/rxxr.shtml) fairly easily.

[...]

PR #242, opened 2021-06-11
Add `no-super-linear-backtracking` rule

RunDevelopment:
This implements one of the solutions from #159. It's a very simple and limited detection method but it should be a good start. However, the limitations of the detection method allow it to fix some of the reported cases which is a valuable feature.

---

I updated some dependencies with CVEs in one of my projects today. One of them was [this](https://app.snyk.io/vuln/SNYK-JS-GLOBPARENT-1016905). This case of ReDoS could have been automatically detected and even fixed using this rule.

I went through some more ReDoS CVEs on snyk and a lot of them were detectable using this rule. Some even fixable. 

I definitely want this rule for v1.0.

---

I also want to add another rule to detect ReDoS not caused by backtracking after this rule got merged.


ota-meshi (approved):
Amazing!

PR #255, opened 2021-06-15
Change `npm run update` to not run tests

RunDevelopment:
This PR changes `npm run update` to not run the tests anymore.

While writing #254, I had to think through our current workflow. One thing I noticed is that it's very strange that `npm run update` runs all tests. Only a few tests are even affected by the changes `npm run update` makes.

With this PR, `npm run update` now only has a single function. This should make it easier for contributors to understand what it does. It even makes it more pleasant to use for me.


ota-meshi (approved):
LGTM!

PR #256, opened 2021-06-20
Add `canReorder` util method

### LINKED ISSUE #251: Improve `no-dupe-disjunctions` by reordering alternatives
To implement `sort-alternatives`, I had to implement a method to decide whether two alternatives can be reordered.

We can use this method to improve the reporting of `no-dupe-disjunctions`. Right now, superset alternatives are not reported by default (e.g. `/\b(?:Foo|\w+)\b/`). But if the alternatives can be reordered, a superset alternative can be become a subset alternative (e.g. `/\b(?:Foo|\w+)\b/` == `/\b(?:\w+|Foo)\b/`).

RunDevelopment:
That took me too many days to write but it's finally finished.

I moved the `canReorder` function into its own util file. I also added some substantial improvements which already make the `sort-alternatives` rule better. This is the first step to resolve #251.

One major addition is what I called determinism equivalence classes. The idea is that we can freely reorder alternatives that start with different characters (e.g. `/foo|bar/`). However, if two alternatives have a common prefix (e.g. `int|integer`), those two alternatives cannot be reordered. Alternatives that have a  common prefix are considered equal.

---

One remaining problem:

The `canReorder` function actually makes one assumption right now: matching direction. Lookbehinds are handled correctly, that's not the problem. The problem is that a partial pattern might get inserted into a lookbehind. This would change the matching direction and the determinism equivalence classes for those alternatives.

E.g. 
```js
var foo = /a.*|b.*/; // left to right, we can order
var bar = RegExp(`(?<=${foo.source})`) // right to left, we can't reorder `a.*|b.*` !
```

The question is: Are we allowed to assume that the matching direction of alternatives is left to right if those alternatives are not children of a lookaround?

In the overwhelming majority of regexes, this will be the same but there is no guarantee. How do we want to handle this?

I actually added an option to `canReorder` that allows us to ignore matching direction. But when will we use that option?


RunDevelopment:
I added [`jest-snapshot`](https://www.npmjs.com/package/jest-snapshot) to make testing easier for myself. But it seems that the library requires NodeJS >=12. This causes an interesting problem: our plugins support NodeJS 10.13.x but your test setup doesn't. 

Do you know any other good snapshot solutions @ota-meshi?

RunDevelopment:
Maybe I should move `canReorder` into `regexp-ast-analysis`? This would also fix the `jest-snapshot` issue.

ota-meshi:
> I added jest-snapshot to make testing easier for myself. But it seems that the library requires NodeJS >=12. This causes an interesting problem: our plugins support NodeJS 10.13.x but your test setup doesn't.

I would like to drop support for Node 10 with the release of v1.0. I will open a PR to do it.

RunDevelopment:
Quick update: I'm currently preparing a new refa release. v0.9.0 will expose some functionality I copied over into this PR (namely the `getBaseSets` and `decomposeIntoBaseSets` functions). So merging this is blocked until the update.

RunDevelopment:
> One remaining problem:

Regarding the matching direction problem: I think we can just ignore it for now. We have a few other rules that also make this assumption, so I think it's ok to add one more.

The problem is an edge case anyway.

---

I think this PR is ready now.

ota-meshi (approved):
LGTM!

PR #257, opened 2021-06-20
Make `optimal-quantifier-concatenation` consistent

### LINKED ISSUE #252: `optimal-quantifier-concatenation` behaves inconsistently around `?`s
image

This is because of a condition I added to all `aa?`. I forgot to consider `a?a`.

This might not look like a problem but it's just really strange when you see it happen in practice:

image

RunDevelopment:
This fixes #252.


ota-meshi (approved):
LGTM!

PR #259, opened 2021-06-24
Fix incorrect pattern parsing when giving RegExp argument to RegExp constructor

### LINKED ISSUE #258: RegExp constructor doesn't support RegExp arguments
```js
const orig = /^https?:\/\//i;
const clone = new RegExp(orig); // => '^' will always reject because it is preceded by a character  regexp/no-useless-assertions
```

ota-meshi:
fixed #258

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -6,7 +6,7 @@ import { RegExpParser, visitRegExpAST } from "regexpp"
 import { CALL, CONSTRUCT, ReferenceTracker } from "eslint-utils"
 import type { Rule, AST, SourceCode } from "eslint"
 import { parseStringTokens } from "./string-literal-parser"
-import { findVariable, getStringIfConstant } from "./ast-utils"
+import { findVariable, getStaticValue, getStringIfConstant } from "./ast-utils"
 import type { ReadonlyFlags, ToCharSetElement } from "regexp-ast-analysis"
 // eslint-disable-next-line no-restricted-imports -- Implement RegExpContext#toCharSet
 import { toCharSet } from "regexp-ast-analysis"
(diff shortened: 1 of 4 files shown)


lib/utils/index.ts
```diff
@@ -349,25 +350,47 @@ function buildRegexpVisitor(
...
+                              ),
+                              ownsFlags: isStringLiteral(flagsNode),
+                          }
+                        : { flagsString: null, ownsFlags: false }
+                if (patternResult.value instanceof RegExp) {
+                    pattern = patternResult.value.source
+                    if (!flagsNode) {
+                        // If no flag is given, the flag is also cloned.
+                        flagsString = patternResult.value.flags
+                        ownsFlags = true
```
RunDevelopment, line 373: Should that really be `true`?

`true` will be correct if we own the pattern but does `getStaticValue` guarantee that?

---

E.g. how will `no-useless-flag` react to this? The `g` flag is useless for `split` but should this even be reported?

```js
var foo = /a/g;
"abc".split(RegExp(foo));
export default foo;
```

ota-meshi: I think `new RegExp (foo)` can be considered to own the flag. If the flag is wrong, user can easily fix it.

I think it should be report and the user should fix it.

```js
var foo = /a/g;
"abc".split(RegExp(foo, ''));
export default foo;
```

Or, if the clone is wrong, fix it as follow:

```js
var foo = /a/g;
"abc".split(foo);
export default foo;
```

RunDevelopment: You're right. I had the wrong idea of what it means to "own flags". I interpreted it as "`RegExp(foo)` own the `g` flag of `/a/g`" and not "`RegExp(foo)` own *its copy* of the `g` flag of `/a/g`".

RunDevelopment (approved):

PR #260, opened 2021-06-24
Drop supports Node.js v10.x and v13.x

ota-meshi:


ota-meshi:
@RunDevelopment 
We are planning v1.0 in the next release, so I think we can merge if this PR is okay.

RunDevelopment:
@ota-meshi We also have a few other breaking PRs. Maybe we could release one last v0.x version right now and then merge all breaking PRs? 

We had a few fixes and additions since v0.12.0, so we might as well make them available to our users before we start breaking things.

ota-meshi:
OK. I will release 0.13 this week or next week as the last v0 release.

README.md
```diff
@@ -37,7 +37,7 @@ npm install --save-dev eslint eslint-plugin-regexp
 > **Requirements**
 >
 > - ESLint v6.0.0 and above
-> - Node.js v8.10.0 and above
+> - Node.js v12.x, v14.x and above
```
RunDevelopment, line 40: It might be nice to write that down in [`package.json`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines) as well.

RunDevelopment (changes requested):

README.md
ota-meshi: Good idea! I added it.

RunDevelopment (approved):

RunDevelopment:
@ota-meshi You think we can merge this?

ota-meshi:
@RunDevelopment Yes. I think we can merge this PR.

PR #264, opened 2021-07-02
Fixed `sort-alternatives` number regex

RunDevelopment:
#262 pointed out a bug in one of the regexes used in `sort-alternatives`.

The problem is obvious once you see it. The fix is also very simple.


ota-meshi (approved):
Thank you!

PR #269, opened 2021-07-06
Fixed `toCharSet` ignoring optional flags

### LINKED ISSUE #265: `no-useless-flag` false positive for `i` flag and Unicode property escapes
See the following example:

```js
var re1 = /\p{Ll}/iu;
```

image

This report is incorrect as `\p{Ll}` matches all lower-case letters. The `i` flag is very much necessary.

RunDevelopment:
This fixes #265.


ota-meshi (approved):
Thank you!

PR #275, opened 2021-07-16
Fix `regexp/negation` false positives

### LINKED ISSUE #266: `regexp/negation` false positives because of `i` flag
See the following example:

```js
var re2 = /[^\P{Ll}]/iu;
```

image

Ignoring the strange suggested fix (this will be a separate issue), the report is incorrect. 

`/[^\P{Ll}]/iu` != `/\p{Ll}/iu`: 
The Canonicalize operation of the `i` flag is applied _before_ the character class is inverted. This example, `\P{Ll}` means "all character that are not lower-case letters". Since`\p{Ll}` contains all upper-case letters, it will also contain all lower-case letters after the Canonicalize operation (not quite right but almost), so `\P{Ll}` matches (almost) all characters. The character class will then match (almost) no character because it is inverted. This is obviously very different from `/\p{Ll}/iu` which matches (almost) all letter characters.

RunDevelopment:
Fixes #266.

### DIFF
--- a/lib/rules/negation.ts
+++ b/lib/rules/negation.ts
@@ -31,23 +31,50 @@ export default createRule("negation", {
             node,
             getRegexpLocation,
             fixReplaceNode,
+            toCharSet,
+            flags,
         }: RegExpContext): RegExpVisitor.Handlers {
             return {
                 onCharacterClassEnter(ccNode) {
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #276, opened 2021-07-18
Add fix and suggestion for `no-unused-capturing-group`

### LINKED ISSUE #230: Add auto-fix for `no-unused-capturing-group`?
Let's insert `?:`?

RunDevelopment:
This adds fixes and suggestions for the `no-unused-capturing-group` rule as discussed #230.

The rule will now always add suggestions where possible. 

Fixes will only be added if the new `fixable` option is set to `true`. The `fixable` options default to `false`, so fixing is opt-in. This option should be a decent compromise.

I also improved the error messages a little.

---

This fixes #230.


ota-meshi (approved):
LGTM! Thank you!

PR #279, opened 2021-07-24
Improve `optimal-quantifier-concatenation`

### LINKED ISSUE #278: Improve `optimal-quantifier-concatenation`
There is a whole class of fixable problems `optimal-quantifier-concatenation` (OQC) can't detect yet. Right now, OQC can only detect (and fix) the concatenation of single-character quantifier (e.g. `\d*\w*`). It's can't detect anything else.

OQC should also be able to detect (and fix) `(?:foo)?\w+` == `\w+`. 

More generally, if we have a single-character quantifier `A{n,∞}` and a quantifier `B{p,q}`, we will be able to simplify `B{p,q}` to `B{p}` if all possible consumed characters of `B` are a subset of `A` (`getConsumedCharacters(B) ⊆ getConsumedCharacters(A)`).

RunDevelopment:
This resolves #278.

This PR includes 2 things:

1. Improved handling of quantifiers as described in #278.
2. Support for nested quantifiers. In some case, we can simplify (or remove) a nested quantifier because of adjacent quantifiers (e.g. `(?:\d+|foo)\w+` -> `(?:\d|foo)\w+`). This is essentially a generalization of the approach we already had.

### DIFF
--- a/lib/rules/optimal-quantifier-concatenation.ts
+++ b/lib/rules/optimal-quantifier-concatenation.ts
@@ -1,3 +1,5 @@
+// eslint-disable-next-line eslint-comments/disable-enable-pair -- x
+/* eslint-disable complexity -- x */
 import type { RegExpVisitor } from "regexpp/visitor"
 import type {
     Alternative,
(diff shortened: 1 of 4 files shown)


ota-meshi (approved):
LGTM!

Issue #231, opened 2021-06-04
`order-in-character-class`: Reported on references


fisker:
```js
const jsxWhitespaceChars = " \n\r\t";
const matchJsxWhitespaceRegex = new RegExp("([" + jsxWhitespaceChars + "]+)");
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ The problem is reported here
```

The location is miss leading, because there is no `\n` there.

Is it hard? I didn't read code, but I think strings are concated, lost original node info?

Another problem `\n`/`\t`/` ` should be escaped in message, they are invisible.

fisker:
Or maybe we can just display the Regexp literal value in message.

fisker:
`no-useless-non-capturing-group` rule have the same issue.

PR #282, opened 2021-08-08
Improve pattern source information

RunDevelopment:
This PR changes how rules determine the position of RegExpp nodes in the original JS source code.

Instead of analyzing the AST in each call of `getRegexpRange`, we now store the information about how the pattern string is constructed in a `PatternSource` instead.

## `PatternSource`

A `PatternSource` is essentially just an array of all parts (concatenated) that make up the pattern string.  Examples:

- `/foo/i`: the instance is `[/foo/i]`
- `RegExp(/foo/i, "m")`: the instance is `[/foo/i]`
- `RegExp("foo")`: the instance is `["foo"]`
- `RegExp("foo" + "bar")`: the instance is `["foo", "bar"]`

Each segment that makes up the pattern has information about its AST node, its string value, and its range within the complete pattern string.

Here is a complete instance for the code:

```js
const space = "\t\n\r ";
new RegExp("([" + space + "]+)");
```

image

Since we know exactly what AST node contributed in what way to the final pattern string, we pinpoint the location of RegExpp AST nodes. This is great for error messages but it also allows us to fix a lot more.

## Comparison

Old:
(nothing is fixable)

image

New:
(everything is fixable)

image

## Semantic differences

The old `getRegexpRange` had 2 porpuses:

1. Get AST ranges for error reports.
2. Get AST ranges for fixes.

The new API by `PatternSoruce` explicitly differentiates these 2 use cases. 

This is important because AST ranges for fixes have to be exact and they have to map to exactly one AST node that we can change. However, AST ranges for error reports don't have to be exact, they just have to be good enough for humans.

The new API:

1. `PatternSoruce#getAstRange` for error reports. This implements a best-effort approach to provide an AST range that is useful to humans.
2. `PatternSoruce#getReplaceRange` for fixes. The returned `PatternReplaceRange` contains all the information to edit source code. 
    `PatternReplaceRange` also has methods to create fixes. These methods will automatically handle escaping (if necessary) and should be used instead of `fixer.replaceTextRange`.

By differentiating these to use cases, we can provide exact ranges for fixes and good approximate ranges for error reporting. This should improve the user experience a lot.

## Rule changes

The above-mentioned semantic changes broke a few rules. All rules now use the new API.

This is quite nice because it means that all rules now benefit from `PatternSource`. This improvement can be seen in the test cases. A few test cases changed because they can now be fixed. The report range also changed on 1 test case.

---

This fixes half of #231.

### DIFF
--- a/lib/rules/control-character-escape.ts
+++ b/lib/rules/control-character-escape.ts
@@ -8,8 +8,8 @@ import {
     CP_TAB,
     createRule,
     defineRegexpVisitor,
-    isRegexpLiteral,
 } from "../utils"
+import { isRegexpLiteral } from "../utils/ast-utils/utils"
 
 const CONTROL_CHARS = new Map<number, string>([
     [0, "\\0"],
--- a/lib/rules/match-any.ts
+++ b/lib/rules/match-any.ts
@@ -2,7 +2,8 @@ import type { RegExpVisitor } from "regexpp/visitor"
 import type { Rule } from "eslint"
 import type { CharacterClass, Node as RegExpNode } from "regexpp/ast"
 import type { RegExpContext } from "../utils"
-import { createRule, defineRegexpVisitor, isRegexpLiteral } from "../utils"
+import { createRule, defineRegexpVisitor } from "../utils"
+import { isRegexpLiteral } from "../utils/ast-utils/utils"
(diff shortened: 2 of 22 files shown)


ota-meshi (approved):
LGTM! Thank you for lot of work!

PR #285, opened 2021-08-10
Improves reporting of the `g` and `y` flags in `regexp/no-useless-flag` rule.

### LINKED ISSUE #253: `no-useless-flag` needs better error messages
image

> The 'g' flag is unnecessary because not using global testing.

The current error message doesn't really say anything. A better error message would be: "The 'g' flag is unnecessary because `String.prototype.split` ignores the 'g' flag." or something like that.

Same problem for other string methods and flags.

Also, now that I got the error message: What's "global testing" @ota-meshi?

---

I wound how we could fix this.

The current implementation seems to mark flags as used. Maybe we could keep a list of everything that uses the regex? If we know that the regex is used by exactly one method and that method doesn't use a flag, we can give a more specific error message. Would that work?

ota-meshi:
This PR improves the reporting of the `g` and `y` flags in `regexp/no-useless-flag` rule.

- Improved the report message to make it easier to see how the g and y flags are used incorrectly.
- Improved the rule to autofix when a regex instance is used only once and given directly to a method.

close #253

### DIFF
--- a/lib/rules/no-useless-flag.ts
+++ b/lib/rules/no-useless-flag.ts
@@ -34,7 +34,19 @@ type CodePathId = {
     loopNode: Statement | undefined
 }
 
+type RegExpMethodKind =
+    | "search"
+    | "split"
+    | "exec"
+    | "test"
+    | "match"
+    | "matchAll"
+    | "replace"
+    | "replaceAll"
+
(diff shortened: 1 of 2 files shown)


RunDevelopment (approved):
Excellent work!

PR #289, opened 2021-08-10
Add `mention` function

### LINKED ISSUE #231: `order-in-character-class`: Reported on references

RunDevelopment:
This fixes #231.

This adds 2 new util functions `mention` and `mentionChar`. These functions have the goal of escaping the raw of RegExp AST nodes. 

The escaping follows similar rules as the [`RegExp.prototype.source` escaping](https://tc39.es/ecma262/multipage/text-processing.html#sec-escaperegexppattern). The rules are:

- Escape `\r` and `\n` just like `RegExp.prototype.source`.
- Don't escape tabs.
- Escape all other control characters as hexadecimal.

This escaping is necessary so we don't have awkward line breaks in error messages and so we don't send control characters to the console/terminal.

Right now, I only used these new functions in a few rules. 

@ota-meshi Should we use them within the entire code base? I think we should. What about you?

### DIFF
--- a/lib/rules/control-character-escape.ts
+++ b/lib/rules/control-character-escape.ts
@@ -10,6 +10,7 @@ import {
     defineRegexpVisitor,
 } from "../utils"
 import { isRegexpLiteral } from "../utils/ast-utils/utils"
+import { mentionChar } from "../utils/mention"
 
 const CONTROL_CHARS = new Map<number, string>([
     [0, "\\0"],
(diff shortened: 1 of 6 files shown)


ota-meshi:
> Should we use them within the entire code base? I think we should. What about you?

I agree with you.
I think we need to add `mention()` to the message that displays the characters.

ota-meshi (approved):

PR #292, opened 2021-08-11
Add `strictTypes` option to `regexp/no-useless-flag` rule

### LINKED ISSUE #205: `no-useless-flag` missing many cases
I noticed we use [`typeTracer.isString`](https://github.com/ota-meshi/eslint-plugin-regexp/blob/1c7b3b58cd6df77e8060037382e251248f28485d/lib/rules/no-useless-flag.ts#L390) to make sure `.split` is used on string, but it can't get the type in many cases, can we just assume all `.split` calls are safe to report? Or at least under an option.

ota-meshi:
This PR adds the `strictTypes` option to the `regexp/no-useless-flag` rule.
The user can loosen the type check by specifying `false` for `strictTypes`.

close #205

### DIFF
--- a/docs/rules/no-useless-flag.md
+++ b/docs/rules/no-useless-flag.md
@@ -157,13 +157,16 @@ No other flags will be checked.
 {
   "regexp/no-useless-flag": ["error",
     {
-      "ignore": [] // An array of "i", "m", "s", "g" and "y".
+      "ignore": [], // An array of "i", "m", "s", "g" and "y".
+      "strictTypes": true
     }
   ]
 }
 ```
 
 - `ignore` ... An array of flags to ignore from the check.
+- `strictTypes` ... If `true`, strictly check the type of object to determine if the regex instance was used in `search()` and `split()`. Default is `true`. This option is only effective for verifying the `g` and `y` flags.  
+  You cannot off of this option if you are using TypeScript.
 
 ### `"ignore": ["s", "g"]`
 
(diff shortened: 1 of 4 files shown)


docs/rules/no-useless-flag.md
```diff
@@ -157,13 +157,16 @@ No other flags will be checked.
...
+      "ignore": [], // An array of "i", "m", "s", "g" and "y".
+      "strictTypes": true
     }
   ]
 }
 ```
 
 - `ignore` ... An array of flags to ignore from the check.
+- `strictTypes` ... If `true`, strictly check the type of object to determine if the regex instance was used in `search()` and `split()`. Default is `true`. This option is only effective for verifying the `g` and `y` flags.  
+  You cannot off of this option if you are using TypeScript.
```
RunDevelopment, line 169:
```suggestion
  This option is always on when using TypeScript.
```

lib/rules/no-useless-flag.ts
```diff
@@ -758,6 +772,32 @@ function createRegExpReferenceExtractVisitor(
...
 }
 
+/**
+ * Parse option
+ */
+function parseOption(
+    userOption: {
+        ignore?: ("i" | "m" | "s" | "g" | "y")[]
+        strictTypes?: boolean
+    } | void,
```
RunDevelopment, line 782: Kinda the same but different.

```suggestion
    } | undefined,
```

RunDevelopment (changes requested):
2 minor things. Apart from that, it's good to go.

docs/rules/no-useless-flag.md
ota-meshi: Thank you!

RunDevelopment (approved):

PR #294, opened 2021-08-11
Shrink report range of `no-useless-non-capturing-group`

RunDevelopment:
This implements [my suggestion](https://github.com/ota-meshi/eslint-plugin-regexp/pull/49#issuecomment-896682733) for `no-useless-non-capturing-group`.

image


ota-meshi (approved):
LGTM!

PR #295, opened 2021-08-11
Use `mention` if necessary

### LINKED ISSUE #291: Change rules to use `mention` function.
https://github.com/ota-meshi/eslint-plugin-regexp/pull/289#issue-707704284

We need to know which rules need to be changed.

RunDevelopment:
This resolves #291.

I only used `mention` where necessary. I didn't use it where I know that the displayed string can't contain control characters.


ota-meshi (approved):
Thank you very much!

PR #297, opened 2021-08-12
Improve `sort-alternatives`

RunDevelopment:
This improves `sort-alternatives` in 2 ways:

1) I added a few more punctuation characters to the list of allowed characters. This will allow the rule for sort more alternatives.

2) The rule can now sort slices. A slice is a continuous section of alternatives (basically the result of calling `alternatives.slice(start, end)`).

   One of the problems of this rule was that adding even one alternative containing forbidden characters meant that the whole group couldn't be sorted anymore. This has now been addressed using slices. Instead of trying to sort the group, the rule will now try to find slices within the group and sort those.

    This has the nice consequence that you can now add alternatives with forbidden characters to a group and most of the group will still be sortable.

   image

### DIFF
--- a/lib/rules/sort-alternatives.ts
+++ b/lib/rules/sort-alternatives.ts
@@ -3,6 +3,10 @@ import type { Alternative, Element, Pattern } from "regexpp/ast"
 import type { RegExpContext } from "../utils"
 import {
     CP_MINUS,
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
Great!

PR #300, opened 2021-08-13
Remove unnecessary `plugins` option for the recommended config in README

chalkygames123:
The `plugins` option can be removed from the example since it is already included in the sharable configs. But please feel free to close this if it is intentionally left in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ npm install --save-dev eslint eslint-plugin-regexp
 <!--USAGE_SECTION_START-->
 
 Add `regexp` to the plugins section of your `.eslintrc` configuration file (you can omit the `eslint-plugin-` prefix)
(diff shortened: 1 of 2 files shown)


RunDevelopment:
I wonder. The `plugins` option isn't necessary but it's also not harmful.

I might be very slightly in favor of keeping things as is. With the `plugins` option, users can easily remove the recommended/all config and use the `rules` option without keeping in mind that they then have to add `"regexp"` to `plugins`. This might make it easier to people less familiar with ESLint to try out/configure the plugin?

What do you think @ota-meshi?

ota-meshi:
If we remove the description of `plugins`, I think we need to go into more detail in the "Advanced Configuration" section.
I think it's still better before it's changed in this PR, so I'll close this PR. Thank you.

PR #305, opened 2021-08-14
Add more playground examples

### LINKED ISSUE #147: Added more Playground examples
We have so many good features but we don't show them to people. We should change that!

Add more examples to the playground. The examples should highlight cool rules/rule combinations and show how this plugin finds real issues.

RunDevelopment:
Fixes #147.


ota-meshi (approved):
Thank you!

PR #306, opened 2021-08-14
Avoid double reporting + flags API improvements

### LINKED ISSUE #304: Avoid double reporting for owned RegExp literals
image

The image speaks for itself. 

The problem gets reported once for the RegExp literal and once for the RegExp constructor. 

---

Solution: 
Since #282, we know exactly which RegExp literals are owned by a `RegExp` constructor. This information is necessary for `PatternSource` because it can only apply fixes to owned RegExp literals.

RunDevelopment:
Fixes #304.

---

This turned out to be more difficult than I thought. Avoiding double reporting was easy enough (you can see that in the first commit). But changing `no-useless-flag` wasn't.

### Changes

Double reporting is now fixed:
image

As for `no-useless-flag`:

- Added a new category of useless flags: The flags of regexp literals that are only used by source. E.g. `/foo/i.source`, `RegExp(/foo/i, '')`

     image

- Improved `getFlagsLocation`. The function now has access to the flags argument directly which makes the whole thing easier to implement.
- Added `getFlagLocation`. This replaces the logic `no-useless-flag` formerly implemented in a private function. It's error locations are more precise because it has access to more information.
 
    Old:
    image

    New:
    image

### DIFF
--- a/lib/rules/no-useless-flag.ts
+++ b/lib/rules/no-useless-flag.ts
@@ -1,5 +1,11 @@
-import type { RegExpContext } from "../utils"
-import { compositingVisitors, createRule, defineRegexpVisitor } from "../utils"
+import type { RegExpContext, RegExpContextForSource } from "../utils"
+import {
+    getFlagsRange,
+    compositingVisitors,
+    createRule,
+    defineRegexpVisitor,
+    getFlagsLocation,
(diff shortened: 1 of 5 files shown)


ota-meshi (approved):
Wonderful!

PR #308, opened 2021-08-15
Add test for ESLint v8

ota-meshi:
If eslint v8 is released early, we will change this PR to without draft.
However, if the eslint v8 support PR for typescript-eslint and vscode-eslint is released early, I will consider closing this PR and opening a PR that modifies devDependencies.

### DIFF
--- a/.github/workflows/NodeCI.yml
+++ b/.github/workflows/NodeCI.yml
@@ -44,7 +44,7 @@ jobs:
                   node-version: 12.13.x
             - name: Install Target Packages
               run: |+
-                  npm install
+                  npm ci
                   npm i -D eslint@6
             - name: Test
               run: npm test
@@ -55,7 +55,7 @@ jobs:
             - uses: actions/setup-node@v2
             - name: Install Target Packages
               run: |+
-                  npm install
+                  npm ci
                   npm i -D eslint@^8.0.0-0
             - name: Test
               run: npm test
--- a/.github/workflows/NodeCI.yml
+++ b/.github/workflows/NodeCI.yml
@@ -49,6 +49,18 @@ jobs:
                   npm install
             - name: Test
               run: npm test
+    test-with-eslint8:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/NodeCI.yml
```diff
@@ -49,6 +49,18 @@ jobs:
...
+    test-with-eslint8:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v2
+            - name: Install Target Packages
+              run: |+
+                  npm i -D eslint@^8.0.0-0
+                  npx rimraf node_modules
+                  npm install
```
RunDevelopment, line 61: What effect does this have? (Same for ESLint v6.x.) FWIK, this should be a no-op.

ota-meshi: It was a workaround to make it work when package-lock.json does not exist. It certainly may not be necessary now😅

ota-meshi: I changed it and it seems to work fine 😄

.github/workflows/NodeCI.yml
```diff
@@ -44,9 +44,19 @@ jobs:
                   node-version: 12.13.x
             - name: Install Target Packages
               run: |+
+                  npm install
```
RunDevelopment, line 47: One small nit. Let's use `npm ci`, it's faster. (Same with v8.x)

```suggestion
                  npm ci
```

ota-meshi: Thank you for pointing out 😅

ota-meshi:
@RunDevelopment Are you using VS Code for your editor? Or are you using another editor?
If you are using any other editor, we will need to consider changes to support this repository for ESLint v8.

I opened PR to vscode-eslint.

https://github.com/microsoft/vscode-eslint/pull/1317

RunDevelopment:
I am using VSCode but thank you for the warning.

However, it's interesting that v8 makes such substantial changes to even break editor integration. Let's hope that this kind of thing doesn't happen too often in the future.

ota-meshi:
Move to #340.

PR #313, opened 2021-08-31
Fix "ignolowercasere" to "lowercase"

regseb:


RunDevelopment:
Seems like something went wrong while copypasting. Thank you for the fix!

RunDevelopment (approved):

PR #315, opened 2021-09-04
Update to `regexp-ast-analysis@0.3.0`

RunDevelopment:
I just released a [new version](https://github.com/RunDevelopment/regexp-ast-analysis/releases/tag/v0.3.0) with a few new features.

Some new features:
- Caching. `regexp-ast-analysis` now has a simple caching system. I'll make a PR for this.
- It now exposes more internal functions. I need those to implement #309.

PR #316, opened 2021-09-04
Improve `no-useless-assertion` for nested assertions

RunDevelopment:
This is a branch of #315.

I noticed that `no-useless-assertion` could handle nested assertions (e.g. `/(?=a|$)b/u`). I am now able to fix this issue because `regexp-ast-analysis` exposes the function I needed to fix this.


ota-meshi (approved):
LGTM!

PR #317, opened 2021-09-04
Use `regexp-ast-analysis`'s caching

RunDevelopment:
This is a branch of #315.

`regexp-ast-analysis` (RAA) finally implements a [caching system](https://rundevelopment.github.io/regexp-ast-analysis/docs/latest/interfaces/cache.html). Since the cache implements the `ReadonlyFlags` interface, it's really easy for us to use the cache. We just have to swap out the plain flags objects we currently use with a cache.

Right now, only RAA's `toCharSet` and `getFirst{Consumed,}Char{After,}` functions are cached. More functions may follow in the future.

**Other changes:**
Cache instances not only implement `ReadonlyFlags` (all fields optional) but even `Required<ReadonlyFlags>`. I took this chance to change the type of `flags` in `RegExpContext` to `Required<ReadonlyFlags>`. This means that we don't have to worry about flags being `undefined` anymore.

---

We added `RegExpContext.toCharSet` because RAA's `toCharSet` used to be uncached. After this PR, `RegExpContext.toCharSet` will just call RAA's `toCharSet` under the hood and do nothing more. Maybe we should remove it?

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -8,7 +8,7 @@ import type { Rule, AST, SourceCode } from "eslint"
 import { getStringIfConstant } from "./ast-utils"
 import type { ReadonlyFlags, ToCharSetElement } from "regexp-ast-analysis"
 // eslint-disable-next-line no-restricted-imports -- Implement RegExpContext#toCharSet
-import { toCharSet } from "regexp-ast-analysis"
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
LGTM!

PR #319, opened 2021-09-04
Add `no-contradiction-with-assertion` rule

### LINKED ISSUE #309: Report and fix confusing quantifiers around assertions
**Description**
Assertions can be used to force certain quantifiers (or alternatives) to be entered. 

E.g. in `a\b-?a`, the `\b` forces `-?` to be entered. The quantifier is useless in this case. 
`a\b-?a` ==<sub>this issue</sub> `a\b-a` ==<sub>no-useless-assertions</sub> `a-a`.

We need to check all quantifiers with a minimum of 0 that are next to an assertion. If the assertion forces us to enter the quantifier, then the minimum should be 1 instead.

---

I think this should be part of `regexp/confusing-quantifier` but I'm not certain. What do you think @ota-meshi?

RunDevelopment:
This resolves #309.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ The rules with the following star :star: are included in the `plugin:regexp/reco
 
 | Rule ID | Description |    |
 |:--------|:------------|:---|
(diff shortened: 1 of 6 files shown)


ota-meshi (approved):
Amazing work! Thank you!

PR #320, opened 2021-09-05
Improve `sort-alternative`'s comparison function

### LINKED ISSUE #307: Improve `sort-alternative`'s comparison function
While testing out the rule on Prism, I noticed one problem that should be fixed: the compare function.

Right now, alternatives are compared 1) by their first consumed character set and then 2) by the raw string value of their source code. The string comparison is a necessary fallback but it's inferior to character-set-based comparisons.

Here are a few examples of incorrect orderings produced by the current comparison function:

- `green|gr[ae]y`
- `(?:script|source)_foo|sample`

The fix is actually pretty simple: Go deeper. Use more character sets for the comparison. The `canReorder` function already implements a way to extract the longest knowable prefix of character sets. Simply factor that logic out into its own class.

---

I classified this as a bug because the `sort-alternative` is supposed to sort lists by byte order. Since it's not doing that correctly right now, it's a bug.

RunDevelopment:
This fixes #307.

This PR has a few improvements:

1. `getLongestPrefix` is now cached.
2. `getLongestPrefix` can now handle groups better. It will interpret `(?:abc|def)` as `[ad][be][cf]`. Not correct per-se but useful for our use case.
3. `sort-alternatives` will now look at the whole prefix before falling back to the string-based comparison.
4. The CharSet-based comparison is better now. I implemented a proper comparison method instead of just comparing the first character in the sets.

### DIFF
--- a/lib/rules/sort-alternatives.ts
+++ b/lib/rules/sort-alternatives.ts
@@ -22,6 +22,7 @@ import type { CharSet } from "refa"
 import { JS } from "refa"
 import { canReorder } from "../utils/reorder-alternatives"
 import { getPossiblyConsumedChar } from "../utils/regexp-ast"
+import { getLongestPrefix } from "../utils/regexp-ast/alternative-prefix"
 
 interface AllowedChars {
     allowed: CharSet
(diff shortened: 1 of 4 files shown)


ota-meshi (approved):
LGTM!

PR #321, opened 2021-09-05
Fix `mergeOverlappingSets` implementation

RunDevelopment:
This fixes a bug in `mergeOverlappingSets`. The function returned more elements than it was supposed to.

I took this chance and reimplemented the function using a simpler and a lot more efficient approach. 

Further, I changed the function's interface to take and return arrays instead of sets. This make a lot of copying and array-to-set conversions unnecessary.

---

This is a branch of #320 because the new test case only reproduces the bug I fixed with the new and improved `getLongestPrefix`.


ota-meshi (approved):
LGTM!

PR #323, opened 2021-09-07
Use `regexp-ast-analysis`'s `toCharSet` everywhere

### LINKED ISSUE #322: Which `toCharSet` should we use going forward?
The reason we added `RegexpContext.toCharSet` was performance (see #148). Since #317, this is no longer a problem. All character-related `regexp-ast-analysis` functions are now cached automatically.

This leaves us in a kinda awkward situation: Which `toCharSet` should we use?

Also, shouldn't we remove [this rule](https://github.com/ota-meshi/eslint-plugin-regexp/blob/750abbc41b66e6ae49fb8531bf7997129254cdba/.eslintrc.js#L45)?

@ota-meshi

RunDevelopment:
This PR removes `RegexpContext.toCharSet` and the associated `ToCharSet` fn type.

A lot of functions only needed a `RegexpContext` parameter to parse characters. I changed these functions to require a `ReadonlyFlags` parameter instead.

---

This resolves #322.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,21 +34,6 @@ module.exports = {
 
         "no-shadow": "off", // ts bug?
         "@typescript-eslint/no-shadow": "error",
-
-        // Rules for implementing this plugin.
-        "no-restricted-imports": [
-            "error",
-            {
-                paths: [
-                    {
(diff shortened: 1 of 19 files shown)


ota-meshi (approved):
Thank you!

PR #324, opened 2021-09-08
Add prefer-lookaround rule

### LINKED ISSUE #183: Rule proposal: Prefer lookahead and lookbehind
Avoid capture unneeded groups.

## Fail

```js
'I love unicorn! I hate unicorn?'.replace(/(?<before>love )unicorn(?<after>!)/, '$<before>🦄$<after>');
```

```js
'I love unicorn! I hate unicorn?'.replace(/(love )unicorn(!)/, '$1🦄$2');
```

## Pass

```js
'I love unicorn! I hate unicorn?'.replace(/(?<=love )unicorn(?=!)/, '🦄');
```

Original posted in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1113, can we implement it here?

ota-meshi:
This PR adds `prefer-lookaround` rule.

There is a problem that the behavior may be different if you replace it with "lookaround" [as you said](https://github.com/ota-meshi/eslint-plugin-regexp/issues/183#issuecomment-828366576). To avoid this problem, I implemented to exclude complicated patterns from the check.

~~This rule is only checked if the pattern consists only of Character, CharacterClass, CharacterSet, CapturingGroup, Group, and Quantifiers max <= 10.
It also controls not to report, when the start capturing group, end capturing group, and the patterns between them include those that can match each pattern.
I think the implementation of this rule can have a lot of false negatives, but I think the rule is useful as far as I can think of at this point.~~

---

close #183


RunDevelopment (changes requested):
I will do a full review soon. Here are a few minor things until then.

lib/rules/prefer-lookaround.ts
```diff
@@ -0,0 +1,633 @@
...
+    extractExpressionReferences,
+    isKnownMethodCall,
+} from "../utils/ast-utils"
+import type { PatternReplaceRange } from "../utils/ast-utils/pattern-source"
+import type { Expression, Literal } from "estree"
+import type { Rule } from "eslint"
+import { mention } from "../utils/mention"
+import { CharSetTreeRoot } from "../utils/char-set-tree"
+
+type ReplaceReference = { ref: string | number; range?: [number, number] }
```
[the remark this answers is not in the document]
ota-meshi: A range is given only if a string literal is given directly to the argument. If there is a range, it may be possible to autofix.

```js
'JavaScript.'.replace(/Java(Script)/g, 'Type$1')
```

However, if the argument string is given indirectly, no range is given. This will not be autofix.

```js
const replacement = 'Type'+'$1'
'JavaScript.'.replace(/Java(Script)/g, replacement)
```

lib/utils/char-set-tree.ts
```diff
@@ -0,0 +1,357 @@
...
+ * }
+ */
+export class CharSetTree {
+    private readonly element: ToCharSetElement
+
+    public readonly charSet: CharSet
+
+    private readonly baseNextList: CharSetTree[]
+
+    private cacheNext: CharSetTree[] | null = null
```
RunDevelopment, line 80:
```suggestion
    private cacheNext: readonly CharSetTree[] | null = null
```

lib/utils/char-set-tree.ts
```diff
@@ -0,0 +1,357 @@
...
+    ): CharSetTreeRoot | null {
+        return CharSetTreeRoot.fromElements([element], regexpContext)
+    }
+
+    /**
+     * Get the CharSetTreeRoot from the given elements. Returns null if it is not a simple pattern.
+     */
+    public static fromElements(
+        baseElements: Element[],
+        regexpContext: RegExpContext,
```
RunDevelopment, line 217: You only ever use the flags of this, so just require a `flags: ReadonlyFlags`.

lib/utils/char-set-tree.ts
```diff
@@ -0,0 +1,357 @@
...
+                element,
+                charSet: toCharSet(element, regexpContext.flags),
+            })
+        }
+    }
+
+    /**
+     * Checks whether the given tree is included.
+     */
+    public includes(target: CharSetTreeRoot): boolean {
```
RunDevelopment, line 347: Please explain this operation to me. I get how this works for simple cases but I don't see how it generalizes to alternations and quantifiers. What does it mean for one tree to _include_ another?

RunDevelopment (changes requested):
Sorry for the wait.

I have a few questions about the general approach you take:

1. What does `CharSetTreeRoot.includes` check? I don't understand which property we check here.
2. Can `CharSetTree`s grow exponentially in size? Let's say I have `(?:a|a|a|a){10}`. Will I get 4<sup>10</sup>=1M `CharSetTree` instances?

I also have a suggestion for another criterion for lookbehinds:

Let's say we have a capturing group _g_ that is a candidate to become a lookbehind (it's the first capturing group, replace string starting with `$1`, etc). Let _c_ be [the union of all characters that can be consumed](https://github.com/ota-meshi/eslint-plugin-regexp/blob/4b01aa57c078abe29b25832f57ccee268c7ad030/lib/utils/regexp-ast/index.ts#L87) by _g_. We now go through all single-character elements (no quantifiers, groups, etc) after _g_. If any of those characters is disjoint with _c_, then we can replace _g_ with a lookbehind.

Example: `"".replace(/(a)Java/g, '$1foo')` (this isn't replaceable with the current implementation.)

ota-meshi:
Thank you for your review. I found that my method didn't work.
The method you suggested is very simple and works well! I have updated this PR to use that method.

lib/rules/prefer-lookaround.ts
```diff
@@ -0,0 +1,625 @@
...
+    start: CapturingGroup | undefined,
+    others: Element[],
+    end: CapturingGroup | undefined,
+    regexpContext: RegExpContext,
+): Set<SideEffect> {
+    const result = new Set<SideEffect>()
+    if (start && others.length) {
+        if (!hasDisjoint(start, ...others)) {
+            result.add(SideEffect.startRef)
+        }
```
RunDevelopment, line 107: This is incorrect. Example: 
```js
> "aaaab".replace(/(a+)a+b/, "$1c")
'aaac'
> "aaaab".replace(/(?<=a+)a+b/, "c")
'ac'
```

We have to be really conservative when checking the elements after the capturing group. If we find an element that does not have a constant length (hint: [`getLengthRange`](https://rundevelopment.github.io/regexp-ast-analysis/docs/latest/modules.html#getlengthrange)), we have to stop (actually, we can get one additional character using `getFirstConsumedCharAfter`).

The function for determining disjointness should look something like this:

```js
const { char } = getPossiblyConsumedChar(start, flags)
for (const element of elements) {
  if (isConstantLength(element)) {
    if (char.isDisjointWith(getPossiblyConsumedChar(element, flags).char)) {
      return true;
    }
  } else {
    const look = FirstConsumedChars.toLook(getFirstConsumedCharPlusAfter(element, "ltr", flags));
    return char.isDisjointWith(look.char)
  }
}
return false
```

(See [`getFirstConsumedCharPlusAfter`](https://github.com/ota-meshi/eslint-plugin-regexp/blob/6e070304c6426b161cb61017ccd30945bdc7365f/lib/utils/regexp-ast/alternative-prefix.ts#L253))

lib/rules/prefer-lookaround.ts
```diff
@@ -0,0 +1,625 @@
...
+        if (!hasDisjoint(end, ...others)) {
+            result.add(SideEffect.endRef)
+        }
+    }
+    if (start && end) {
+        if (!hasDisjoint(start, end)) {
+            result.add(SideEffect.startRef)
+            result.add(SideEffect.endRef)
+        }
+    }
```
RunDevelopment, line 119: Frankly, I am not sure whether this is correct. The method I proposed was intended for the start capturing group.

I _believe_ this could work but I will have to think about this some more. I'll write back later.

RunDevelopment: I thought for a bit and it doesn't work.

```js
> "abaaaba".replace(/ab(a+)/g, "c$1")
'caaaba'
> "abaaaba".replace(/ab(?=a+)/g, "c")
'caaca'
```

As for what I think does work:

A quick reminder before: What property are we trying to verify?
We are trying the verify that the capturing group (CG) behaves in the same way as a lookahead. This is only the case if the end CG does not consume characters that the regex can consume in the next match.

### Without `g` flag
Therefore we can trivially convert the end CG into a lookahead if the `g` flag isn't present because there won't be a "next match".

### With `g` flag
If the `g` flag is present, then this whole thing is more complex. I have 2 conditions: a simple with many false negatives and a complex one.

#### The simple one
We can replace the end CG with a lookahead if the set of all consumed characters of the end CG is disjoint with the first character consumed by the regex.

While this condition guarantees correctness, it's also very conservative and will likely produce false negatives. 

#### The complex one

Now a little diagram. I want to illustrate the possible ways in which the lookahead of the end CG can overlap with the next match. I will call the range asserted by the lookahead of the end CG asserted _E_ and the next match _R_. There are 3 possible configurations in which _E_ and the next match of _R_ can be:

```
1) Disjoint
<~~E~~>
        <--R-->

2) R in E
<~~E~~>
 <-R->

3) E and R overlap but R is not completely in E
<~~E~~>
   <-R->
```

We need case 1 to do the conversion, so we just have to show that cases 2 and 3 are not possible.

Let _X_ be the language accepted by the end CG and let _Y_ be the language accepted by the whole regex. (I use _•_ for concatenation.)

Case 2 is not possible if _X ∩ (Σ* • Y • Σ*) = ∅_.

Let _Z_ be the set of all non-empty prefixes of _Y_. Case 3 is not possible if _X ∩ (Σ* • Z) = ∅_.

Both conditions can be implemented using refa's NFAs.

#### Which one?

Frankly, I don't know whether the performance cost of converting the whole pattern into an NFA is worth it. 

I'll let you decide which one to implement @ota-meshi.

---

Note: Now the conditions for the start and end capturing group are fully independent.

ota-meshi: Thanks for the detailed explanation!!
I don't understand the `complex one` yet, but I think a simple implementation is probably fine.
I think that even a simple implementation will work pretty well for users who want to use this rule.

ota-meshi: As I was working, I noticed that the following results were different.

```js
"abaaaba".replace(/(a)ba+/g, "$1c")
// 'acba'
"abaaaba".replace(/(?<=a)ba+/g, "c")
// 'acc'
```

This means that the converted result of `"".replace(/(a)Java/g, '$1foo')` will change.

```js
"aJavaJava".replace(/(a)Java/g, '$1foo')
// 'afooJava'
"aJavaJava".replace(/(?<=a)Java/g, 'foo')
// 'afoofoo'
```

We probably need to check the last element, whether or not the last element is a capturing group.

ota-meshi:
I changed this PR. Please check it again.

lib/rules/prefer-lookaround.ts
```diff
@@ -0,0 +1,659 @@
...
+        result.add(SideEffect.startRef)
+        result.add(SideEffect.endRef)
+    }
+    return result
+
+    /** Checks whether the given target element has disjoint in elements.  */
+    function hasDisjoint(target: Element, elements: Element[]) {
+        const chars = getPossiblyConsumedChar(target, flags)
+        for (const element of elements) {
+            if (isConstantLength(element)) {
```
RunDevelopment, line 134: Sorry, I made a mistake. We also have to make sure that the element consumes at least one character.

Assertions have a constant length of 0 but don't consume any characters, `getPossiblyConsumedChar(assertion)` is the empty set and the empty set is disjoint with everything.

lib/rules/prefer-lookaround.ts
RunDevelopment: You're right. We need to check that the first character of the start group is disjoint with the last character of the whole pattern.

lib/rules/prefer-lookaround.ts
```diff
@@ -0,0 +1,659 @@
...
+    }
+
+    const last = end || others[others.length - 1]
+    const first = start || others[0]
+    const lastChars = getPossiblyConsumedChar(last, flags)
+    const firstChar = getFirstConsumedChar(first, "ltr", flags)
+    if (!firstChar.char.isDisjointWith(lastChars.char)) {
+        result.add(SideEffect.startRef)
+        result.add(SideEffect.endRef)
+    }
```
RunDevelopment, line 127: What exactly does this check?

RunDevelopment (changes requested):

lib/rules/prefer-lookaround.ts
RunDevelopment: You're right. We need to check that the first character of the start group is disjoint with the last character of the whole pattern.

lib/rules/prefer-lookaround.ts
ota-meshi: I intended to implement `The simple one`. Did I make a mistake in using the regexp-ast-analysis API?

lib/rules/prefer-lookaround.ts
ota-meshi: Thank you for letting me know! I will fix it.

lib/rules/prefer-lookaround.ts
RunDevelopment: I was confused by `last`. The condition I came up with only has meaning if there is an end capturing group.

This is more what I meant.

```ts
if (end) {
	const first = start || others[0]
	if (first) {
		const { char } = getPossiblyConsumedChar(end, flags)

		const firstChar = FirstConsumedChars.toLook(
			getFirstConsumedCharPlusAfter(first, "ltr", flags),
		)
		if (!firstChar.char.isDisjointWith(lastChars.char)) {
			result.add(SideEffect.endRef)
		}
	}
}
```

ota-meshi:
I fixed the remaining issue. But `coveralls` seem to be down😅. I will need to re-run later.

RunDevelopment (approved):
LGTM;

PR #327, opened 2021-09-13
Fix `control-character-escape` false negatives

### LINKED ISSUE #326: `control-character-escape` doesn't report control characters in `/re/.source` expressions
**Description**
As the title says, `control-character-escape` doesn't report control characters in `re.source` expressions if those expressions are owned by a `RegExp` constructor.

image

RunDevelopment:
fixes #326.

I used `PatternSource`'s replace ranges to determine whether the control character is part of a RegExp literal or not.

image


ota-meshi (approved):
LGTM! tahnk you!

PR #328, opened 2021-09-17
Improve `prefer-lookaround`

RunDevelopment:
I made a PR to address my review comments. I adjusted all conditions at least a little bit. Now everything should just work, I think.

However, there is a bug I don't know how to fix: It doesn't fix `"aaaaaa".replace(/(^a+)a(a)/, "$1b$2")` to `"aaaaaa".replace(/(^a+)a(?=a)/, "$1b")`. It does fulfill all conditions and we do get an error message but it's not fixable for some reason.


ota-meshi (approved):
Thank you! I'll look into the remaining issues.

Issue #325, opened 2021-09-08
Bring over all of ESLint's regex rules


RunDevelopment:
ESLint has a few really good rules for regexes (e.g. [`no-misleading-character-class`](https://eslint.org/docs/rules/no-misleading-character-class)). We have already brought over some of them and I wonder whether we should bring over the rest as well? We have more precise error reporting (better user experience) and can handle complex `RegExp` constructors better (more error reports), so our users would benefit from that.

Consensus: Yes, let's bring them all over.

- [x] [`no-control-regex`](https://eslint.org/docs/rules/no-control-regex)
  #333
- [x] [`no-div-regex`](https://eslint.org/docs/rules/no-div-regex) (not necessary)
- [x] [`no-empty-character-class`](https://eslint.org/docs/rules/no-empty-character-class)  
  #329
- [x] [`no-invalid-regexp`](https://eslint.org/docs/rules/no-invalid-regexp)
- [x] [`no-misleading-character-class`](https://eslint.org/docs/rules/no-misleading-character-class) 
  #332
- [x] [`no-regex-spaces`](https://eslint.org/docs/rules/no-regex-spaces) (subset of `regexp/perfer-quantifier`)
- [x] [`no-useless-backreference`](https://eslint.org/docs/rules/no-useless-backreference)
- [x] [`prefer-named-capture-group`](https://eslint.org/docs/rules/prefer-named-capture-group)
  #330 
- [x] [`prefer-regex-literals`](https://eslint.org/docs/rules/prefer-regex-literals) (?)
- [x] [`require-unicode-regexp`](https://eslint.org/docs/rules/require-unicode-regexp)
  #331 

Regarding `prefer-regex-literals`: I don't think that we can improve this rule, so we don't need to bring it over.

ota-meshi:
I think it's good to implement the regex related rules of ESLint as the rules of this plugin.
I think it benefits users, as you say.

ota-meshi:
`no-div-regex` rule reports regex that looks like division. It may be possible to improve the reporting position, but I think it is also possible with the ESLint's rule. So I don't think it's necessary to move to this plugin.

RunDevelopment:
I want to do `no-misleading-character-class`. I'll make it fixable using the string semantics from [this proposal](https://github.com/tc39/proposal-regexp-set-notation#whats-the-match-order-for-character-classes-containing-strings).

RunDevelopment:
[...]

PR #329, opened 2021-09-18
Add `regexp/no-empty-character-class` rule

ota-meshi:
Related to #325

### DIFF
--- a/docs/rules/no-empty-character-class.md
+++ b/docs/rules/no-empty-character-class.md
@@ -2,11 +2,11 @@
 pageClass: "rule-details"
 sidebarDepth: 0
 title: "regexp/no-empty-character-class"
-description: "disallow character classes that does not match all characters"
+description: "disallow character classes that match no characters"
 ---
 # regexp/no-empty-character-class
 
-> disallow character classes that does not match all characters
+> disallow character classes that match no characters
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
 
--- a/docs/rules/no-empty-character-class.md
+++ b/docs/rules/no-empty-character-class.md
@@ -2,19 +2,21 @@
 pageClass: "rule-details"
 sidebarDepth: 0
 title: "regexp/no-empty-character-class"
-description: "disallow empty character classes"
+description: "disallow character classes that does not match all characters"
 ---
 # regexp/no-empty-character-class
 
-> disallow empty character classes
+> disallow character classes that does not match all characters
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
 
 ## :book: Rule Details
 
-This rule disallows empty character classes.
+This rule reports character classes that cannot match all characters.
 
-This rule inspired by [no-empty-character-class] rule. The position of the report is improved over the core rule, and the argument of `new RegExp()` is also checked.
+The character class that cannot match all characters is an empty character class and a negative character class that contains all characters.
+
+The reports for this rule include reports for the ESLint core [no-empty-character-class] rule. That is, if you use this rule, you can turn off the ESLint core [no-empty-character-class] rule.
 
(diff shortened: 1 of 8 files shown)


RunDevelopment:
I was wondering whether it might be useful to [evaluate](https://rundevelopment.github.io/regexp-ast-analysis/docs/latest/modules.html#matchesnocharacters) the character class? 

This does depart from what ESLint does but it might be useful I think. I imagine this rule as an opposite to `match-any`.

ota-meshi:
That's good idea. Do you think we need to change the rule name?

RunDevelopment:
You mean into e.g. `match-none`? I don't think so. While the rule does deviate from ESLint's implementation a little, the intent is still the same. We just check more thoroughly.

ota-meshi:
Thank you for your opinion. I have updated this rule.

lib/rules/no-empty-character-class.ts
```diff
@@ -1,19 +1,23 @@
...
             // TODO Switch to recommended in the major version.
             // recommended: true,
             recommended: false,
         },
         schema: [],
         messages: {
-            unexpected: "Unexpected empty character class.",
+            empty:
+                "Since this character class is empty, it cannot match all characters.",
+            cannotMatchAll: "This character class cannot match all characters.",
```
RunDevelopment, line 20:
```suggestion
            cannotMatchAny: "This character class cannot match any characters.",
```

lib/rules/no-empty-character-class.ts
```diff
@@ -1,19 +1,23 @@
...
 import type { RegExpVisitor } from "regexpp/visitor"
 import type { RegExpContext } from "../utils"
 import { createRule, defineRegexpVisitor } from "../utils"
 
 export default createRule("no-empty-character-class", {
     meta: {
         docs: {
-            description: "disallow empty character classes",
+            description:
+                "disallow character classes that does not match all characters",
```
RunDevelopment, line 10:
```suggestion
                "disallow character classes that match no characters",
```

docs/rules/no-empty-character-class.md
```diff
@@ -2,19 +2,21 @@
...
 
-> disallow empty character classes
+> disallow character classes that does not match all characters
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
 
 ## :book: Rule Details
 
-This rule disallows empty character classes.
+This rule reports character classes that cannot match all characters.
```
RunDevelopment, line 15:
```suggestion
This rule reports character classes that cannot match any characters.
```

docs/rules/no-empty-character-class.md
```diff
@@ -0,0 +1,52 @@
...
+
+> disallow character classes that does not match all characters
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule reports character classes that cannot match all characters.
+
+The character class that cannot match all characters is an empty character class and a negative character class that contains all characters.
```
RunDevelopment, line 17:
```suggestion
Character classes that cannot match any characters are either empty or negated character classes of elements that contain all characters.
```

lib/rules/no-empty-character-class.ts
```diff
@@ -0,0 +1,52 @@
...
+            category: "Possible Errors",
+            // TODO Switch to recommended in the major version.
+            // recommended: true,
+            recommended: false,
+        },
+        schema: [],
+        messages: {
+            empty:
+                "Since this character class is empty, it cannot match all characters.",
```
RunDevelopment, line 19:
```suggestion
                "This character class matches no characters because it is empty.",
```

RunDevelopment (changes requested):
There was some confusion around "any" and "all".

ota-meshi:
Thank you for the review! I have changed this PR.

RunDevelopment (approved):

PR #331, opened 2021-09-19
Add `regexp/require-unicode-regexp` rule

RunDevelopment:
relates to #325.

I also made some changes to the `RegExpContext`.

- `getFlagsLocation`: If a regex literals doesn't have any flags, it will now point to the last `/` instead of the characters after the last `/`.
- `buildRegexpVisitor`: It will now try to dereference owned variables given as the flags argument. This means that we will be able to fix more flag reports.
- `fixReplaceFlags`: If there is no flags argument given to a `RegExp` constructor (and the first argument is a string), then trying to add flags will add a flags string. E.g. `RegExp("foo")` -> `RegExp("foo", "u")`
- I changed the semantics of `RegExpContext.flagsString === null`. It used to mean "no flag argument OR unable to evaluate flag argument". It now only means "unable to evaluate flag argument". This semantic is more useful and necessary for the correct of `require-unicode-regexp`.


ota-meshi (approved):
LGTM!

PR #332, opened 2021-09-20
Add `regexp/no-misleading-unicode-character` rule

RunDevelopment:
Relates to #325.

This adds the `no-misleading-unicode-character` rule as discussed [here](https://github.com/ota-meshi/eslint-plugin-regexp/issues/325#issuecomment-922271310). There are a few differences to ESLint's `no-misleading-character-class` rule:

- We support misleading quantifiers.
- Reports are auto-fixed.
- Reports are much more granular.
- We ignore characters using escape sequences. We only care about visible characters, not hexadecimal escapes. E.g. ESLint's rule will report `/[\ud83d\udc4d]/`, we won't.

One open question: Should we auto-fix?
Right now, the rule will auto-fix wherever possible using groups. We essentially assume that people want to treat graphemes as single units. That is a very reasonable assumption but an assumption nonetheless.

### DIFF
--- a/lib/rules/no-misleading-unicode-character.ts
+++ b/lib/rules/no-misleading-unicode-character.ts
@@ -1,0 +1,276 @@
+import type { RegExpVisitor } from "regexpp/visitor"
+import type { RegExpContext } from "../utils"
+import { isEscapeSequence, createRule, defineRegexpVisitor } from "../utils"
+import GraphemeSplitter from "grapheme-splitter"
+import type { ReadonlyFlags } from "regexp-ast-analysis"
+import { mention, mentionChar } from "../utils/mention"
+import type {
+    CharacterClass,
+    CharacterClassElement,
+    Quantifier,
+} from "regexpp/ast"
+import type { PatternRange } from "../utils/ast-utils/pattern-source"
+
+const splitter = new GraphemeSplitter()
+
+/** Returns whether the given string starts with a valid surrogate pair. */
+function startsWithSurrogate(s: string): boolean {
+    if (s.length < 2) {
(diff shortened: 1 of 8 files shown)


lib/rules/no-misleading-unicode-character.ts
```diff
@@ -0,0 +1,276 @@
...
+    const prefix = problems
+        .map((p) => p.grapheme)
+        .sort((a, b) => b.length - a.length)
+        .join("|")
+
+    // The rest of the character class
+    let ccRaw = cc.raw
+    for (let i = problems.length - 1; i >= 0; i--) {
+        const { start, end } = problems[i]
+        ccRaw = ccRaw.slice(0, start) + ccRaw.slice(end)
```
ota-meshi, line 142: Maybe we need to do the following:

```suggestion
        ccRaw = ccRaw.slice(0, start - cc.start) + ccRaw.slice(end - cc.start)
```

I found that it couldn't be autofixed well in the following cases:

```js
/[竈門禰󠄀豆子]|[煉󠄁獄杏寿郎]/
```

ota-meshi (changes requested):

ota-meshi:
> Should we auto-fix?

I personally find auto-fix useful for me, but I think it may be better for some language native users not to auto-fix.
I think it may be better to move to the Suggestion API to avoid confusion. Or add `fixable` option like `no-unused-capturing-group`.

RunDevelopment:
I like the `fixable` option.

ota-meshi (approved):
LGTM! Thank you!

PR #333, opened 2021-09-22
Add `regexp/no-control-character` rule

RunDevelopment:
Relates to #325.

Changes to ESLint's rule:

- There are suggestions for the characters `\0\r\n\t\v\f`.
- This rule allows `\0\r\n\t\v\f` in `RegExp` constructor strings. E.g. `RegExp("\n")` is allowed.
- The rule detects escaped control characters. E.g. `RegExp("\\\x0a")` is reported.


ota-meshi (approved):
LGTM!

>  RegExp("\n") is allowed.

That's nice! Thank you!

PR #335, opened 2021-09-24
Fix `no-useless-lazy` false negative

### LINKED ISSUE #334: `no-useless-lazy` doesn't handle negated lookarounds correctly
**Description**
I just noticed that `no-useless-lazy` didn't report `/\b\w+?\b/` even though the lazy modifier is useless because the quantifier is effectively possessive.

image

```js
/* ✗ BAD */
var foo = /f\w+?(?=\W)/; // correct
var foo = /f\w+?\b/;     // false negative
var foo = /f\w+?(?!\w)/; // false negative
var foo = /f\w+?$/;      // false negative
```

RunDevelopment:
This fixes #334.

image

### DIFF
--- a/lib/rules/no-useless-lazy.ts
+++ b/lib/rules/no-useless-lazy.ts
@@ -102,10 +102,7 @@ export default createRule("no-useless-lazy", {
                             matchingDir,
                             flags,
                         )
-                        if (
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
LGTM!

PR #338, opened 2021-09-26
Improve `sort-flags` to fix unknown patterns

### LINKED ISSUE #337: `sort-flags` should autofix flags that's a string literal
**Information:**
- ESLint version: 7.32.0
- `eslint-plugin-regexp` version: 1.3.0

**Description**

`/a/us` is fixable, but `RegExp('a', 'us')` is not.

RunDevelopment:
Fixes #337.

The main change is that `fixReplaceFlags` now takes an `includePattern` parameter. This controls whether a flags fix includes the whole pattern as well. Including the whole pattern is necessary in pretty much all cases except for sorting (and maybe some others), so it will default to `true`.

### DIFF
--- a/lib/rules/sort-flags.ts
+++ b/lib/rules/sort-flags.ts
@@ -42,7 +42,7 @@ export default createRule("sort-flags", {
                         loc: getFlagsLocation(),
                         messageId: "sortFlags",
                         data: { flags: flagsString, sortedFlags },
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
Thank you for the quick fix! LGTM!

RunDevelopment:
Thank you for the quick release @ota-meshi!

PR #339, opened 2021-09-27
Chore: minor upgrade dev deps

ota-meshi:
This PR is a minor upgrade of devDependencies.

In addition, I apply the format and autofix, manually fixed that cannot be autofix, and fixed the Type error.

Also, the versions of some packages below have been changed from `^` to `~`:

- eslint-plugin-regexp
- prettier
- typescript

This helps to fix format and type errors little by little, as renovate sends a lot of PR. (We can close it PR if we don't need it.)


lib/utils/mention.ts
```diff
@@ -39,14 +39,14 @@ function escape(value: string): string {
     //  2. Escape all non-escape control character
     return (
         value
-            .replace(/\\([\s\S])/g, (m, char) => {
+            .replace(/\\(?<char>[\s\S])/gu, (m, _1, _o, _s, { char }) => {
```
RunDevelopment, line 42: Not sure if that's an improvement...

Just because we gave a group name, doesn't mean that we are required to use it.
```suggestion
            .replace(/\\(?<char>[\s\S])/gu, (m, char) => {
```

lib/utils/mention.ts
```diff
@@ -39,14 +39,14 @@ function escape(value: string): string {
...
                 if (char.charCodeAt(0) < 0x20) {
                     return escapeControl(char)
                 }
                 return m
             })
-            // eslint-disable-next-line no-control-regex -- x
-            .replace(/[\0-\x1f]/g, escapeControl)
+            // eslint-disable-next-line no-control-regex, regexp/no-control-character -- x
```
RunDevelopment, line 48: Why is `regexp/no-control-character` active? Are we using our `all` config?

lib/utils/type-tracker/index.ts
```diff
@@ -587,6 +592,7 @@ export function createTypeTracker(context: Rule.RuleContext): TypeTracker {
         const declarations = symbol && symbol.declarations
         const declaration = declarations && declarations[0]
         if (
+            declaration &&
```
RunDevelopment, line 595: This doesn't look like reformating. There are also a few other changes of this sort in `lib/utils/type-tracker/**` files. What's that about?

tools/update-docs.ts
```diff
@@ -154,9 +157,9 @@ This rule was introduced in eslint-plugin-regexp ${this.since}
         const { meta } = this.rule
 
         this.content = this.content.replace(
-            /<eslint-code-block(.*?)>/gu,
-            (_t, str) => {
-                const ps = str
+            /<eslint-code-block(?<attrs>.*?)>/gu,
+            (_t, _1, _o, _s, { attrs }) => {
```
RunDevelopment, line 161: Same here.
```suggestion
            (_t, attrs) => {
```

tools/update-readme.ts
```diff
@@ -36,14 +36,14 @@ fs.writeFileSync(
         .replace(/<!--DOCS_IGNORE_START-->[\s\S]*?<!--DOCS_IGNORE_END-->/gu, "")
         .replace(
             // eslint-disable-next-line regexp/no-super-linear-backtracking -- it's acceptable here
-            /\(https:\/\/ota-meshi.github.io\/eslint-plugin-regexp(.*?)([^/]*\.html)?\)/gu,
-            (_ptn, c1: string, c2: string) => {
-                let result = `(.${c1}`
-                if (c2) {
+            /\(https:\/\/ota-meshi.github.io\/eslint-plugin-regexp(?<paths>.*?)(?<name>[^/]*\.html)?\)/gu,
+            (_ptn, _1, _2, _o, _s, { paths, name }) => {
```
RunDevelopment, line 40: And here.
```suggestion
            (_ptn, paths, name) => {
```

tools/update-rules.ts
```diff
@@ -7,7 +7,9 @@ import { rules } from "./lib/load-rules"
  * Convert text to camelCase
  */
 function camelCase(str: string) {
-    return str.replace(/[-_](\w)/gu, (_, c) => (c ? c.toUpperCase() : ""))
+    return str.replace(/[-_](?<char>\w)/gu, (_, _1, _o, _s, { char }) =>
```
RunDevelopment, line 10: And here
```suggestion
    return str.replace(/[-_](?<char>\w)/gu, (_, char) =>
```

RunDevelopment (changes requested):
I agree with this PR in general. Just a few nits.

lib/utils/mention.ts
ota-meshi: In the current rule config, that code is reported by `regexp/no-unused-capturing-group`.

```js
/* eslint
   regexp/no-unused-capturing-group: error,
   regexp/prefer-named-capture-group: error
*/
"value"
  .replace(/\\(?<char>[\s\S])/gu, (m, _1, _o, _s, { char }) => {
    // ...
  });

// Capturing group 'char' has a name, but its name is never used. ( regexp/no-unused-capturing-group )
"value"
  .replace(/\\(?<char>[\s\S])/gu, (m, char) => {
    // ...
  });

// Capture group '([\s\S])' should be converted to a named or non-capturing group. ( regexp/prefer-named-capture-group )
"value"
  .replace(/\\([\s\S])/gu, (m, char) => {
    // ...
  });
```

[DEMO](https://ota-meshi.github.io/eslint-plugin-regexp/playground/#eJydkcFOhDAQhl/lTy/ABmi8ou4efASP1pgujEDCtmRKNyaEd7dFd10PZtUmk04n/0y/v51FbRsSlZAbkBt6MykDgKmlt1EaW3jjHTVFrcfJc2/aomXrxwrEbDm/FI9Mr8SF0Yeznr6pldlIZZQ46sGTErG3ZBoHXVMqVVjp7q7uNG+flHJKPT5nsvU50kOOl5sQNoTLMSOKsGS432JeCSAlyrKM+ZLdhjtMKDyckLFCIIltCTrtoBEpc+z9hH5y6wl92OlIjGi4RHr1EZD9000s/46eTuzpeUoC11k/NNgTamsC8EQNJvtpqoFlGGu+YD9GXBj6+aOuOfqzFZEL9gM5Uc3L8g6Ey8sa)

We may need to add an option to one of the rules to allow you to opt out of reporting the replace with function.

What do you think of adding the `ignoreUnusedNameWhenReplaceWithFunction` option, which excludes reports when the replace with function even if the name is not used, to the `regexp/no-unused-capturing-group` rule?

lib/utils/type-tracker/index.ts
RunDevelopment: Ok, that's fine.

lib/utils/mention.ts
RunDevelopment: Isn't the reason we have this problem in the first place that `no-unused-capturing-group` does 2 things? It:

1. checks whether capturing groups are used (as the name suggests) and
2. checks whether the names of capturing groups are used.

Couldn't we add an option `checkNames: boolean` instead? This option would control if names are checked.

lib/utils/mention.ts
RunDevelopment: I agree that using `recommended` is better.

lib/utils/mention.ts
ota-meshi: I also considered the option of not checking the use of all names, but I think it's more readable to use names in replaces with placeholders.

`"str".replace(/(?<foo>foo)/, '$<foo>')` > `"str".replace(/(?<foo>foo)/, '$1')`

So I think it's a good ​to add the `ignoreUnusedNameWhenReplaceWithFunction` option.

However, I think your point about the rule checking for things other than what the name suggests is true.
Do you think it's better to do the following?

- Remove the name check from the `no-unused-capturing-group` rule.
- Add new rule `no-unused-named-capturing-group` rule.
- Add an option to exclude replace with function in the new rule.

lib/utils/mention.ts
ota-meshi: I will change this PR once the `no-unused-capturing-group` issue is resolved :+1:

lib/utils/mention.ts
RunDevelopment: Checking whether names are used is useful, so I think that we should keep it in some form.

I think adding a new rule might be the best solution. Since both `no-unused-capturing-group` and the new rule have to know whether a capturing group is used at all, we should also factor out the logic for capturing group usage (as part of `RegExpContext` just like `getUsageOfPattern`). 

This logic can be useful to other rules as well (e.g. `prefer-named-replacement`). (Do we have other rules that need this kind of capturing group information too?)

Also, I think `no-unused-capturing-group-name` is a better name.

> `"str".replace(/(?<foo>foo)/, '$<foo>')` > `"str".replace(/(?<foo>foo)/, '$1')`

Agreed. We should probably have a rule for this. Something like `prefer-named-replacement`.

ota-meshi: The `prefer-named-replacement` rule is sounds good to me!

> Do we have other rules that need this kind of capturing group information too?

We also need to verify the return values of `exec`,` match`, and `matchAll`, but I don't think it's more readable than replacement. I think it's a good to add this as a separate rule. For example `prefer-regexp-result-array-groups`?

RunDevelopment: > We also need to verify the return values of `exec`, `match`, and `matchAll`, but I don't think it's more readable than replacement. I think it's a good to add this as a separate rule. For example `prefer-regexp-result-array-groups`?

I don't understand. [`no-unused-capturing-group`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-unused-capturing-group.html) already verifies `exec` and `match`, doesn't it?

ota-meshi: I may have misunderstood. I will reread English.

RunDevelopment: I think I misunderstood your use of "verfiy". I interpreted your "verify" as "checks that a capturing group is used". 

However, `no-unused-capturing-group` already does this checking, so I don't see why we would need a new rule. What would `prefer-regexp-result-array-groups` do exactly?

ota-meshi: I was misunderstanding what you wrote.

> Do we have other rules that need this kind of capturing group information too?

Currently, I don't think there is anything other than `prefer-named-replacement` rule.

However, I propose again to add the `prefer-regexp-result-array-groups` rule. The `prefer-regexp-result-array-groups` rule uses that information. I also have to give a detailed description of the `prefer-regexp-result-array-groups` rule.

`prefer-regexp-result-array-groups` is a rule that inherits some named capturing group checks from `no-unused-capturing-group`. 
This rule prefers named access over array index access when using named capturing group in regex. Perhaps the process is similar to `prefer-named-replacement`.
When adding `prefer-regexp-result-array-groups`, exclude the same check from `no-unused-capturing-group`.

The example code that `prefer-regexp-result-array-groups` warns is:

```js
const reFoo = /(?<foo>foo)/
let resArr
while (resArr = reFoo.exec(str)) {
  const foo = resArr[1]
  //             ^^^ Use `resArr.groups.foo` instead.
  //             Currently it is warned by no-unused-capturing-group, but change it to not warn.
}
```

The sample code that the new rule does not warn is:

```js
const reFoo = /(?<foo>foo)/
let resArr
while (resArr = reFoo.exec(str)) {
  const foo = resArr.groups.foo // correct
}
```

```js
const reFoo = /(foo)/
let resArr
while (resArr = reFoo.exec(str)) {
  const foo = resArr[1] // Not using named capturing group in regex.
}
```

```js
const reFoo = /(foo)/
let resArr
while (resArr = reFoo.exec(str)) {
  const match = resArr[0] // The capturing group was not used. It's warned by no-unused-capturing-group rule.
}
```

The `prefer-regexp-result-array-groups` rule checks not only the `exec` but also the return values of `match` and `matchAll`.

RunDevelopment: Oh, I see. Thank you for the clarification! That rule sounds really good. I'll make an issue for this, so we can track progress.

---

I think we almost solved the original topic with this. `no-unused-captruing-group` will no longer check name usage. The two proposed rules `prefer-regexp-result-array-groups` and `prefer-named-replacement` will handle it.

But this still leaves cases like the `replace` of this comment. Are we going to:

1. Add a new rule specifically for unused names?
2. Add a new option to `no-unused-captruing-group` that can turn name-checking off?

Some people might not have the `prefer-named-capturing-group` enabled and want to see unused names.

ota-meshi: > Add a new rule specifically for unused names?

I don't think we need to add any rules other than `prefer-regexp-result-array-groups` and `prefer-named-replacement`.
There is no way to report the following code, but I think that's okay.

```js
"value"
  .replace(/\\(?<char>[\s\S])/gu, (m, char) => {
    // ...
  });
```

If a user opens an issue requesting it, we will consider a new rule.

> Add a new option to no-unused-captruing-group that can turn name-checking off?  
> Some people might not have the prefer-named-capturing-group enabled and want to see unused names.

That's a good point. I think it's a good to add a new option and remove it in v2. However, I will remove the report of unused names in the replacer function.

ota-meshi:
I will change this PR to Draft until some issues are resolved and new rules are applied.

lib/utils/mention.ts
RunDevelopment: > If a user opens an issue requesting it, we will consider a new rule.

I think that's a good plan.

> I think it's a good to add a new option and remove it in v2. However, I will remove the report of unused names in the replacer function.

We could also just remove the name-checking completely. The rule will report strictly less, so we can release it in the next minor version.

---

Also, we should definitely remove the name "fixing". Auto-removing unused names is not a good idea when we have rules that use those names. See [here](https://ota-meshi.github.io/eslint-plugin-regexp/playground/#eJwFwVEKgzAMANCrhHypzJX9dmM7iNlHZ2MpSFPSBgTx7nvvxFUio0c3Abc9lw7KiY/qisxWrHGc11C7aS5pTipWPSyErCpKeIMTtnyE384euhrD9YXJUaHihs9rE3mHEUKLQPRw2e6dWx8ICccnXn+VjSmq).

image

However, technically the fault lies with `prefer-named-backreference`. Fixes from `prefer-named-backreference` depend on the capturing group name remaining unchanged, so the capturing group name should be included in its fix (to create conflicts with anyone trying the change the capturing group name).

So how do we fix this? Do we:

1. Allow nobody to change/remove capturing group names.
2. Add the capturing group name to the fix for rules depending on the name.

This is relevant for #347 and #344 as well.

ota-meshi: Thank you for letting me know. I changed my mind.
I think `no-unused-capturing-group` name-checking can be considered a kind of bug that conflicts with other rules and should completely remove name-checking. I will quit adding option.

RunDevelopment (approved):
Thank you @ota-meshi!

RunDevelopment:
Out of curiosity: How many of those `.group` changes were auto-fixed by the rule?

PR #343, opened 2021-09-28
Add `insideCharacterClass` option to `prefer-d`

### LINKED ISSUE #341: `regexp/prefer-d`: Add an option for whole `\d`s
**Description**
I think that the range `0-9` can be more readable than `\d` in some situations. E.g. `[0-9a-f]` vs `[\da-f]`.

I think we should an option that makes it so that only whole character classes that are equal to `\d` or `\D` are reported. So `[0-9]` will be reported but `[0-9a-f]` won't.

What do you think @ota-meshi?

RunDevelopment:
This fixes #341.

The new option adds 3 modes for character class elements:

- `ignore`: This is the don't-care option.
- `d` *(default)*: Replace all with `\d`. This is the current behaviour.
- `range`: Replace all with `0-9`.

### DIFF
--- a/docs/rules/prefer-d.md
+++ b/docs/rules/prefer-d.md
@@ -34,7 +34,79 @@ var foo = /[^0-9]/;
 
 ## :wrench: Options
 
-Nothing.
+```json5
+{
+  "regexp/prefer-d": [
+    "error",
+    {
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #345, opened 2021-09-28
Add `regexp/use-ignore-case` rule

### LINKED ISSUE #201: Suggest `i` flag if it simplifies the pattern
If a pattern can be simplified by adding the `i` flag (e.g. `/[a-fA-F]/` -> `/[a-f]/i`), it might be a good idea to suggest that to the user.

The conditions for this rule to add the `i` flag (and subsequently simplify the pattern) are:

1. The `i` flag isn't present.
2. All characters, character classes, and character sets match the same character(s) regardless of the `i` flag (aka. the flag doesn't change the meaning of the pattern.)
3. There is at least one character class that can be simplified because of the added `i` flag. E.g. a character range can be removed, or a character class can be replaced by a character or character set, or similar.

RunDevelopment:
This resolves #201, a somewhat long-standing issue.

Other changes:

- I moved the case-invariance logic from `no-useless-flags` into its own file, so `use-ignore-case` can use it.
- I fixed a bug with `fixReplaceFlags`. The new `includePattern` parameter added in #338 was ignored for regex literals. I noticed this big because `use-ignore-case`'s fix didn't work.


ota-meshi (approved):
Awesome!

PR #348, opened 2021-09-30
Improve `no-unused-capturing-group` docs

### LINKED ISSUE #346: Further explanation for `no-unused-capturing-group`
I coincidentally found stylelint/eslint-config-stylelint#141 today and I thought that the discussion around `no-unused-capturing-group` was interesting.

In the documentation of `no-unused-capturing-group`, we say that non-capturing groups should be used to "optimize regular expressions" (as in "improve performance") but I don't think that's the only reason non-capturing groups should be preferred. 

Using non-capturing groups also communicates intend more clearly. A capturing group is similar to a variable in that its value (the captured text) is stored and can be accessed, and like with variables, unused capturing groups are code smell. If a regex contains a capturing group, then I'll expect its captured text to be used in some form or another.

If it's alright with you @ota-meshi, then I'll add this reasoning to the documentation.

RunDevelopment:
Fixes #346.

### DIFF
--- a/docs/rules/no-unused-capturing-group.md
+++ b/docs/rules/no-unused-capturing-group.md
@@ -14,7 +14,7 @@ since: "v0.6.0"
 
 ## :book: Rule Details
 
-This rule aims to optimize regular expressions by replacing unused capturing groups with non-capturing groups.
+This rule reports unused capturing groups.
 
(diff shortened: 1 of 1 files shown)


ota-meshi (approved):
Thank you!!

Issue #347, opened 2021-09-30
Add `regexp/prefer-result-array-groups` rule


RunDevelopment:
From [this comment](https://github.com/ota-meshi/eslint-plugin-regexp/pull/339#discussion_r719249323).

> `prefer-regexp-result-array-groups` is a rule that inherits some named capturing group checks from `no-unused-capturing-group`. 
> This rule prefers named access over array index access when using named capturing group in regex. Perhaps the process is similar to `prefer-named-replacement`.
> When adding `prefer-regexp-result-array-groups`, exclude the same check from `no-unused-capturing-group`.
> 
> The example code that `prefer-regexp-result-array-groups` warns is:
> 
> ```js
> const reFoo = /(?<foo>foo)/
> let resArr
> while (resArr = reFoo.exec(str)) {
>   const foo = resArr[1]
>   //                ^^^ Use `resArr.groups.foo` instead.
>   //                Currently it is warned by no-unused-capturing-group, but change it to not warn.
> }
> ```
> 
> The sample code that the new rule does not warn is:
> 
> ```js
> const reFoo = /(?<foo>foo)/
> let resArr
> while (resArr = reFoo.exec(str)) {
>   const foo = resArr.groups.foo // correct
> }
> ```
> 
> ```js
> const reFoo = /(foo)/
> let resArr
> while (resArr = reFoo.exec(str)) {
>   const foo = resArr[1] // Not using named capturing group in regex.
> }
> ```
> 
> ```js
> const reFoo = /(foo)/
> let resArr
> while (resArr = reFoo.exec(str)) {
>   const match = resArr[0] // The capturing group was not used. It's warned by no-unused-capturing-group rule.
> }
> ```
> 
> The `prefer-regexp-result-array-groups` rule checks not only the `exec` but also the return values of `match` and `matchAll`.

RunDevelopment:
I think we should change the name a little. This whole plugin is about regexes, so I think `prefer-result-array-groups` is enough.

ota-meshi:
`prefer-result-array-groups` is good :+1:! `regexp` is in the prefix.

Issue #344, opened 2021-09-28
Add `prefer-named-replacement` rule


RunDevelopment:
As per [this comment](https://github.com/ota-meshi/eslint-plugin-regexp/pull/339#discussion_r717477491).

> `"str".replace(/(?<foo>foo)/, '$<foo>')` > `"str".replace(/(?<foo>foo)/, '$1')`

ota-meshi:
SGTM!

PR #349, opened 2021-10-01
Remove the unused name check from `regexp/no-unused-capturing-group` rule

ota-meshi:
This PR removes unused name checks from the `regexp/no-unused-capturing-group` rule.

Remove the unused name check because it was a check that could not be inferred from the name of this rule and there was a conflict with the `prefer-named-backreference` rule.

See https://github.com/ota-meshi/eslint-plugin-regexp/pull/339#discussion_r720309918

We also plan to add rules that prefers using the names of named capturing groups defined in regex. #344, #347.

---

This PR includes a refactoring to move the capture group usage to common function.
This will also be used in the new rule #344 and #347.

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -1,7 +1,13 @@
 import type * as ESTree from "estree"
 import type { RuleListener, RuleModule, PartialRuleModule } from "../types"
 import type { RegExpVisitor } from "regexpp/visitor"
-import type { Element, Node, Pattern, Quantifier } from "regexpp/ast"
+import type {
+    CapturingGroup,
+    Element,
+    Node,
+    Pattern,
+    Quantifier,
+} from "regexpp/ast"
 import { RegExpParser, visitRegExpAST } from "regexpp"
 import { CALL, CONSTRUCT, ReferenceTracker } from "eslint-utils"
 import type { Rule, AST, SourceCode } from "eslint"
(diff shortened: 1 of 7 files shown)


lib/utils/index.ts
```diff
@@ -79,6 +88,16 @@ type RegExpContextBase = {
      */
     getUsageOfPattern: () => UsageOfPattern
 
+    /**
+     * Returns the capturing group references
+     */
+    getCapturingGroupReferences: () => CapturingGroupReference[]
```
RunDevelopment, line 94: These references don't include which capturing group they are about, correct? Are they in the same order as the `getAllCapturingGroups` array?

If yes, then it might be nice to return a `ReadonlyMap<CapturingGroup, CapturingGroupReference>` instead.

ota-meshi: It currently does not contain information about the regex AST capturing group.
`getAllCapturingGroups` is used to get the count. It is necessary to parse the contents indicated by the argument of the replacer function.

```js
str.replace(/(a)(b)/, (match, p1, p2, offset, string, groups) => {});
str.replace(/(a)(b)(c)/, (match, p1, p2, p3, offset, string, groups) => {});
```

RunDevelopment: Ah, my bad. I didn't think of groups having multiple references.

RunDevelopment (approved):
LGTM; Thank you @ota-meshi!

PR #351, opened 2021-10-02
Add prefer-result-array-groups rule

### LINKED ISSUE #347: Add `regexp/prefer-result-array-groups` rule

ota-meshi:
close #347

### DIFF
--- a/docs/rules/prefer-result-array-groups.md
+++ b/docs/rules/prefer-result-array-groups.md
@@ -1,0 +1,63 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "regexp/prefer-result-array-groups"
+description: "enforce using result array `groups`"
+---
+# regexp/prefer-result-array-groups
+
+> enforce using result array `groups`
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule reports and fixes index access in regexp result array that do not use the name of their referenced capturing group.
+
+<eslint-code-block fix>
+
+```js
+/* eslint regexp/prefer-result-array-groups: "error" */
+
+const regex = /(?<foo>a)(b)c/
+let re
+while (re = regex.exec(str)) {
+    /* ✓ GOOD */
+    var p1 = re.groups.foo
+    var p2 = re[2]
+
+    /* ✗ BAD */
+    var p1 = re[1]
+}
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+```json
+{
+  "regexp/prefer-result-array-groups": ["error", {
+    "strictTypes": true
+  }]
+}
+```
+
+- `strictTypes` ... If `true`, strictly check the type of object to determine if the string instance was used in `match()` and `matchAll()`. Default is `true`.  
+  This option is always on when using TypeScript.
+
+## :couple: Related rules
+
+- [regexp/prefer-named-backreference]
+- [regexp/prefer-named-capture-group]
+- [regexp/prefer-named-replacement]
+
+[regexp/prefer-named-backreference]: ./prefer-named-backreference.md
+[regexp/prefer-named-capture-group]: ./prefer-named-capture-group.md
+[regexp/prefer-named-replacement]: ./prefer-named-replacement.md
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/ota-meshi/eslint-plugin-regexp/blob/master/lib/rules/prefer-result-array-groups.ts)
+- [Test source](https://github.com/ota-meshi/eslint-plugin-regexp/blob/master/tests/lib/rules/prefer-result-array-groups.ts)
--- a/docs/rules/prefer-result-array-groups.md
+++ b/docs/rules/prefer-result-array-groups.md
@@ -13,22 +13,22 @@ description: "enforce using result array `groups`"
 
 ## :book: Rule Details
 
-This rule reports and fixes index access in regexp result array that do not use the name of their referenced capturing group.
+This rule reports and fixes regexp result arrays where named capturing groups are accessed by index instead of using [`groups`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges#using_named_groups).
 
 <eslint-code-block fix>
 
 ```js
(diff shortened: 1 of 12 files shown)


lib/rules/prefer-result-array-groups.ts
```diff
@@ -0,0 +1,104 @@
...
+            {
+                type: "object",
+                properties: {
+                    strictTypes: { type: "boolean" },
+                },
+                additionalProperties: false,
+            },
+        ],
+        messages: {
+            unexpected: "Unexpected indexed access from regexp result array.",
```
RunDevelopment, line 24: It would be nice to have a more descriptive error message.

```suggestion
            unexpected: "Unexpected indexed access for the named capturing group '{{ name }}' from regexp result array.",
```

docs/rules/prefer-result-array-groups.md
```diff
@@ -0,0 +1,63 @@
...
+
+const regex = /(?<foo>a)(b)c/
+let re
+while (re = regex.exec(str)) {
+    /* ✓ GOOD */
+    var p1 = re.groups.foo
+    var p2 = re[2]
+
+    /* ✗ BAD */
+    var p1 = re[1]
```
RunDevelopment, line 31: I know this is really pedantic, but could we rename `re`? People typically use `re` to mean regular expression but this variable holds a regex match. 
```suggestion
let match
while (match = regex.exec(str)) {
    /* ✓ GOOD */
    var p1 = match.groups.foo
    var p2 = match[2]

    /* ✗ BAD */
    var p1 = match[1]
```

docs/rules/prefer-result-array-groups.md
```diff
@@ -0,0 +1,63 @@
...
+## :book: Rule Details
+
+This rule reports and fixes index access in regexp result array that do not use the name of their referenced capturing group.
+
+<eslint-code-block fix>
+
+```js
+/* eslint regexp/prefer-result-array-groups: "error" */
+
+const regex = /(?<foo>a)(b)c/
```
RunDevelopment, line 23: The regex needs a `g` or `y` flag or else the `while` loop is an infinite loop.
```suggestion
const regex = /(?<foo>a)(b)c/g
```

docs/rules/prefer-result-array-groups.md
```diff
@@ -0,0 +1,63 @@
...
+# regexp/prefer-result-array-groups
+
+> enforce using result array `groups`
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule reports and fixes index access in regexp result array that do not use the name of their referenced capturing group.
```
RunDevelopment, line 16: I changed it a little for clarity and added a link to relevant MDN docs.

```suggestion
This rule reports and fixes regexp result arrays where named capturing groups are accessed by index instead of using [`groups`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges#using_named_groups).
```

lib/rules/prefer-result-array-groups.ts
```diff
@@ -0,0 +1,104 @@
...
+                                              return null
+                                          }
+                                          return fixer.replaceTextRange(
+                                              [
+                                                  openingBracket.range![0],
+                                                  memberNode.range![1],
+                                              ],
+                                              `${
+                                                  memberNode.optional ? "" : "."
+                                              }groups.${cgNode.name}`,
```
RunDevelopment, line 88: `groups` is [an optional property in TS](https://github.com/microsoft/TypeScript/blob/main/lib/lib.es2018.regexp.d.ts). This fix will cause a TS compiler error if strict null checks are enabled.

If the current file is a `.ts` file and strict null checks are enabled, then the fix should be `groups!.name`.

Does TS have an API to get the current TS config, so we can adjust the fix?

RunDevelopment (changes requested):
Awesome work! 

Just a few minor nits.

lib/rules/prefer-result-array-groups.ts
ota-meshi: Thank you for letting me know. I didn't consider for TypeScript.
I will change it to give `!` If strict type checking is enabled.

ota-meshi: I changed it to check the type before autofix.
Also, if a simple array type is specified, autofix will be stopped.

```ts
let match: any[] | null
while (match = regex.exec(foo)) {
    const p1 = match[1]
    // ...
}
```

Also, if it is a `any` type, `!` is not inserted.

lib/rules/prefer-result-array-groups.ts
ota-meshi: Thank you!

lib/rules/prefer-result-array-groups.ts
RunDevelopment: Amazing work! Thank you!

RunDevelopment (approved):

PR #403, opened 2022-03-11
Fix false positives for matchAll in `regexp/no-unused-capturing-group` rule

### LINKED ISSUE #399: matchAll: false positive of regexp/no-unused-capturing-group
```javascript
const a = [...'var s1="s1"\nvar s2="s2"'.matchAll(/var (s\d+)=("[^"]+")/g)].map(
  m => [m[1], m[2]]
)
```

ota-meshi:
fixes #399

### DIFF
--- a/lib/utils/extract-capturing-group-references.ts
+++ b/lib/utils/extract-capturing-group-references.ts
@@ -136,6 +136,51 @@ type ExtractCapturingGroupReferencesContext = {
     isString: (node: Expression) => boolean
 }
 
+type ArrayMethodName = Exclude<keyof unknown[], "length" | symbol | number>
+const WILL_KNOWN_ARRAY_METHODS: {
+    [key in ArrayMethodName]: {
+        iteration?: number[]
+        result?: "element" | "array" | "iterator"
+    }
+} = {
+    toString: {},
+    toLocaleString: {},
+    pop: { result: "element" },
+    push: {},
+    concat: { result: "array" },
+    join: {},
+    reverse: { result: "array" },
+    shift: { result: "element" },
+    slice: { result: "array" },
+    sort: { iteration: [0, 1], result: "array" },
+    splice: { result: "array" },
+    unshift: {},
+    indexOf: {},
+    lastIndexOf: {},
+    every: { iteration: [0] },
+    some: { iteration: [0] },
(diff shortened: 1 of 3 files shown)


lib/utils/extract-capturing-group-references.ts
```diff
@@ -638,3 +737,23 @@ function getNamedArrayRef(
...
+        parent.callee !== expression
+    ) {
+        return null
+    }
+    return parent
+}
+
+/** Checks whether the given name is a well known array method name. */
+function isWillKnownArrayMethodName(name: string): name is ArrayMethodName {
+    return Boolean(WILL_KNOWN_ARRAY_METHODS[name as ArrayMethodName])
```
RunDevelopment, line 758: Typo.
```suggestion
function isWellKnownArrayMethodName(name: string): name is ArrayMethodName {
    return Boolean(WELL_KNOWN_ARRAY_METHODS[name as ArrayMethodName])
```

lib/utils/extract-capturing-group-references.ts
```diff
@@ -136,6 +136,51 @@ type ExtractCapturingGroupReferencesContext = {
...
+    toString: {},
+    toLocaleString: {},
+    pop: { result: "element" },
+    push: {},
+    concat: { result: "array" },
+    join: {},
+    reverse: { result: "array" },
+    shift: { result: "element" },
+    slice: { result: "array" },
+    sort: { iteration: [0, 1], result: "array" },
```
RunDevelopment, line 155: `iterator` should probably be renamed. It really is a list of all parameter indexes that will be assigned an array element (if the first argument of the array method is a function).

Maybe `elementParams` or `elementParameters`?

Also, could you please document this on the type? It took me a little while to figure this out.

RunDevelopment (changes requested):
Good work @ota-meshi!

I only have 2 minor nits.

lib/utils/extract-capturing-group-references.ts
ota-meshi: Thanks!

ota-meshi:
@RunDevelopment Thank you for checking this PR!
I have added the changes you requested to this PR. So please check again.

RunDevelopment (approved):

PR #404, opened 2022-03-11
Make `regexp/no-dupe-disjunctions` account for nested alternatives

### LINKED ISSUE #402: `regexp/no-dupe-disjunctions` should "de-sugar" character classes
**Description**
While writing some regexes, I noticed the `regexp/no-dupe-disjunctions` didn't say anything about a regex of this from:

```js
/a+|[abc]/
```

The `a` in `[abc]` is clearly useless but didn't get reported. This is because `regexp/no-dupe-disjunctions` looks at the `[abc]` as a whole and only detects the overlap between `a+` and `[abc]`. It doesn't understand that the overlap is really subset (with `a` in `[abc]`) and therefor you get this report with `report: "all"` even though the problem is trivial.

---

Some test cases:

```js
// bad
var foo = /a+|[abc]/; // remove `a` in `[abc]`
var foo = /a+|a|b|c/; // remove `a` (this currently works)
var foo = /a+|[a-f]/; // change `a-f` to `b-f`

var foo = /a|[ab]a/; // remove `a` in `[ab]`
var foo = /a|aa|bb/; // remove `aa` (this currently works)

// ok
var foo = /c+|[a-f]/;
```

RunDevelopment:
This fixes #402.

---

To implement this, I needed to introduce 2 new concepts to the code base:

1. Nested alternatives: These are alternatives or things that behave like alternatives (e.g. character class elements). They are called nested because they are the (transitive) child nodes of some root alternative. 
    E.g. if `a(c|b)[de]` is the root alternative, then `b`, `c`, `d`, and `e` are nested alternatives.
3. Partial NFAs: The partial NFA of a given root alternative and nested alternative is the NFA that contains all paths of the root alternative that go through the nested alternative.
    E.g. for the root alternative `(a|b)(c|d|e)f` and nested alternative `d`, the language of the partial NFA will be `(a|b)df`.

These 2 concepts are useful because they allow us to ask: Is there some part of an alternative that is a subset? 

This is actually quite a bit more general than simple de-sugaring of character classes. This PR allows use to detect partial subsets and duplications within alternatives, and that includes character classes.

---

Side note: I expected this to massively impact performance. We are potentially creating a bunch more NFAs and DFAs than before, but (surprisingly) I couldn't measure any difference when I ran this PR against Prism's 2k regexes. (It did find 5 bugs though.)


lib/rules/no-dupe-disjunctions.ts
```diff
@@ -121,6 +125,27 @@ function containsAssertionsOrUnknowns(
...
+                throw new Error()
+            },
+            onBackreferenceEnter() {
+                throw new Error()
+            },
+        })
+        return false
+    } catch (error) {
+        return true
+    }
```
ota-meshi, line 146: Is it possible to use hasSomeDescendant?

```suggestion
    return hasSomeDescendant(
        node,
        (d) => d.type === "Assertion" || d.type === "Backreference",
    )
```

lib/rules/no-dupe-disjunctions.ts
```diff
@@ -175,6 +200,90 @@ function toNFA(
...
+                    } else if (min.value + 1 === max.value) {
+                        nested.push(min, max)
+                    } else {
+                        nested.push(charElement, min, max)
+                    }
+                } else {
+                    nested.push(charElement)
+                }
+            }
+            if (nested.length >= 1) yield* nested
```
ota-meshi, line 241: I prefer to use yield directly rather than using an array. 

```suggestion
            for (const charElement of e.elements) {
                if (charElement.type === "CharacterClassRange") {
                    const min = charElement.min
                    const max = charElement.max
                    if (min.value === max.value) {
                        yield charElement
                    } else if (min.value + 1 === max.value) {
                        yield min
                        yield max
                    } else {
                        yield charElement
                        yield min
                        yield max
                    }
                } else {
                    yield charElement
                }
            }
```

ota-meshi (changes requested):
Thank you for your wonderful work!!
I made two comments.

lib/rules/no-dupe-disjunctions.ts
RunDevelopment: Yes, it is! I should have though of that. Thank you.

lib/rules/no-dupe-disjunctions.ts
RunDevelopment: I made a mistake here. The `nested.length >= 1` should have been `nested.length > 1`. I'll fix that.

We can't yield elements directly because we want to ensure that we yield _at least 2_ elements per character class. The same check is also done for array alternatives. 

```ts
if (e.alternatives.length > 1) {
    yield a
}
```

The "at least 2" condition is an optimization. I.e. in `a(b)c`, we don't have to check the `b` alternative because we know that the partial NFA of `b` will be the complete NFA of `a(b)c`, so we don't have to check `b`. 

The array is necessary for character classes because ranges may yield 1-3 elements, so it's not so easy to know how many elements we'll yield in the end.

RunDevelopment: The optimization has been fixed and documented in 5f985a1fc88dcf5fb67ff1fbe09dc2f8d529bafe.

ota-meshi: I got it!

RunDevelopment:
Sorry for the inconvenience. I just found a bug in `PartialParser` that caused it to construct the partial NFA incorrectly for alternatives. I should have tested this as well.

tests/lib/rules/no-dupe-disjunctions.ts
```diff
@@ -391,6 +391,86 @@ tester.run("no-dupe-disjunctions", rule as any, {
...
+                    column: 4,
+                },
+            ],
+        },
+        {
+            code: String(/a|(a|b)a/),
+            errors: [
+                {
+                    message:
+                        "Unexpected useless element. All paths of '(a|b)a' that go through this element are already covered by 'a'. This element can be removed. Careful! This alternative contains capturing groups which might be difficult to remove.",
```
RunDevelopment, line 459: The capturing group warning is incorrect. I'll fix it.

RunDevelopment: Again, sorry for the inconvenience. The warning has been fixed in b42349108f6d7c099413d054eb4d88fa7eea39ec.

RunDevelopment:
Again, sorry for the inconvenience, @ota-meshi. 

I think I'm done. Could you please review?

ota-meshi (approved):
LGTM!

PR #406, opened 2022-03-15
Add suggestions for `regexp/no-dupe-disjunctions`

### LINKED ISSUE #405: Add suggestions to `regexp/no-dupe-disjunctions`
**Description**
It might be useful to have suggestions (but not fixes) for `regexp/no-dupe-disjunctions`.

What do think @ota-meshi?

RunDevelopment:
This fixes #405.

This is a branch of #404. I'll rebase this after #404 is merged.

---

Regarding the changes: It's all relatively simple except for the range replacement. 

I wanted suggestions for ranges (e.g. `a+|[a-z]` -> `a+|[b-z]`), but this wasn't easy. The main problems was: How do we print the new range? We have to print a new character (in the example above, we needed to print `b` to make `b-z`), but do we do that?

I decided to print them as hexadecimal escapes, except when the character is in [`allowedCharacterRanges`](https://ota-meshi.github.io/eslint-plugin-regexp/settings/#available-settings). This is a somewhat unintended use case for this setting, so I hope this won't case problems down the road.


lib/rules/no-dupe-disjunctions.ts
```diff
@@ -1104,6 +1145,103 @@ export default createRule("no-dupe-disjunctions", {
...
+                if (containsCapturingGroup) {
+                    // we can't just remove a capturing group
+                    return []
+                }
+
+                if (result.type === "Overlap" || result.type === "Superset") {
+                    // the types of results cannot be trivially fixed by
+                    // removing an alternative.
+                    return []
+                }
```
ota-meshi, line 1186: I prefer to move this branch to the beginning of the function, as I think the type checking on line 1168 will be more stringent.
What do you think?

RunDevelopment: Good idea!

ota-meshi (approved):
LGTM! Thank you!

PR #415, opened 2022-04-16
Upgrade to RAA v0.4.1

RunDevelopment:
I wanted to move everything related to `canReorder` into my `regexp-ast-analysis` library for a long time now, and I finally got around to doing it.

This PR removes the code I added around `canReorder` and uses the RAA functions instead.

I also put some work into choosing a cleaner abstraction for `getLongestPrefix` as well as improving the function itself. As a result of these improvements, I found a false negative in our tests. `/(?:js|jso?n?)$/` was listed as a valid regex for `no-dupe-disjunctions` even though the `js` alternative can be removed.

While looking through the `no-dupe-disjunctions` tests, I also found that the `yml` alternatives in both `/(?:yml|ya?ml)$/` and `/(?:yml|ya?ml)/` can be removed. I'm currently investigating this.


RunDevelopment:
> I also found that the `yml` alternatives in both `/(?:yml|ya?ml)$/` and `/(?:yml|ya?ml)/` can be removed. I'm currently investigating this.

I made a minor improvement to `canReorder` and it is not able to reorder `/(?:yml|ya?ml)$/`. I'll make a new RAA release tomorrow or the day after.

About `/(?:yml|ya?ml)/`: I have an idea that could work (RunDevelopment/regexp-ast-analysis#2), but I'll have to think about this some more.

RunDevelopment:
I implemented my idea and make a new release (v0.4.1). RAA is now able to prove that `/(?:yml|ya?ml)/` can be reordered. I adjusted the tests of `no-dupe-disjunction` accordingly.

ota-meshi (approved):
Great! Thank you!

PR #419, opened 2022-04-18
Upgrade to RAA v0.5.1 and improve `sort-alternatives`

### LINKED ISSUE #416: The order enforced by `sort-alternatives` depends on characters outside the sorted alternatives
**Description**
Right now, alternatives are sorted 1) by the prefix returned by `getLongestPrefix` and 2) by raw source code. The bug is that `getLongestPrefix` will also possibly return one character after the alternatives, which means that the sort order of the alternatives depends not only on the alternatives but also on the characters after them. This has some surprising consequences. E.g. right now, `sort-alternatives` will sort as follows:

```js
var foo = /^(?:a|ab)a/u;
var foo = /^(?:ab|a)c/u;
```

I would classify this as a bug.

Fixing this issue will be quite easy. The RAA version of `getLongestPrefix` in #415 has options that control its behavior. This includes an option to ensure that the prefix is only of the alternative and does not contain characters from after it.

RunDevelopment:
This fixes #416.

> Fixing this issue will be quite easy.

As it turned out, it was not that easy. Just disabling the `includeAfter` option wasn't good enough, so I added a new option `insideOnly`. This new option is actually what we want, but to implement this option, I needed to change the `getLengthRange` function as well.

Anyway, it all works now and using `getLengthRange` is actually easier now.

### DIFF
--- a/lib/rules/no-useless-assertions.ts
+++ b/lib/rules/no-useless-assertions.ts
@@ -252,7 +252,7 @@ export default createRule("no-useless-assertions", {
                     const range = getLengthRange(assertion.alternatives)
                     // we only check the first character, so it's only correct if the assertion requires only one
                     // character
-                    if (range && range.max === 1) {
+                    if (range.max === 1) {
                         // require exactness
                         if (
                             firstOf.exact &&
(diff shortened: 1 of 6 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #421, opened 2022-04-18
Remove unused code in `sort-alternatives`

RunDevelopment:
I forgot to remove some old code in #320. This code used to be part of the comparison, but was replaced by the prefix-based comparison function we use today.


ota-meshi (approved):
Thank you!

Issue #420, opened 2022-04-18
Change character set comparison function in `sort-alternatives`?


RunDevelopment:
Since we need to be able to sort character sets in `sort-alternatives`, I implemented a comparison function in #320. However, I recently discovered that this function doesn't sort that way you'd expect.

Example: `sort-alternatives` wants to swap the following alternatives:
```js
var re = /\b(?:sample|(?:un)?stable)\b/u;
```

This is quite strange to me because the first characters of `sample` and `(?:un)?stable` are `[s]` and `[su]` respectively, and I think `[s]` should come before the character set `[su]`.

Honestly, I don't know whether this is a bug or not. We don't specify *how* alternatives are sorted, so *technically* any order is correct. What do you think @ota-meshi?

ota-meshi:
> I think [s] should come before the character set [su].

I agree with you :+1:

RunDevelopment:
Resolved by #423.

PR #423, opened 2022-04-20
Use LSS comparison for `sort-alternatives`

RunDevelopment:
This PR implements a char set comparison function that has the properties described in #420. It works by comparing two char sets by comparing the lists of characters accepted by the char sets. At least, that's what it behaves like. It's still implemented based on ranges.

That's what I implemented, but I'm not sure whether this is actually better than what we had. As we can see when looking at the tests, `(?:sample|(?:un)?stable)` is now sorted "correctly". However, there are a few examples where the order is now worse. E.g. `(false|[ft]|true)` used to be ordered as `([ft]|false|true)` which was a lot better IMO.

---

Frankly, I'm starting to question whether sorting based on prefixes is even a good idea. It's (relatively) easy to do, but is it the best we can do?

How about this sorting method: Let `a` and `b` be the two alternatives we want to compare. Let `w` be the lexicographically smallest string that is accepted by either `a` or `b` but not both. If no such string exists, then `a == b`. If `w` is accepted by `a`, then `a < b`. If `w` is accepted by `b`, then `a > b`.

This sorting method will neatly order `(?:sample|(?:un)?stable)` and `([ft]|false|true)`. It's also more intuitive IMO.

The only problem with this sorting method is that it doesn't work. The "lexicographically smallest string" is only defined for finite sets of strings. I.e. What's the lexicographically smallest string for `a+b`? `ab` > `aab` > `aaab` > `aaaaaaaaaaaab` > ... I'm not sure how this could be fixed.

### DIFF
--- a/lib/rules/sort-alternatives.ts
+++ b/lib/rules/sort-alternatives.ts
@@ -96,39 +96,46 @@ function compareByteOrder(a: string, b: string): number {
  * Compare two char sets by byte order.
  */
 function compareCharSets(a: CharSet, b: CharSet): number {
-    // empty char set > everything else
-    if (a.isEmpty) {
-        return 1
-    } else if (b.isEmpty) {
-        return -1
+    // The basic idea here is the following:
+    // We want to sort the two sets based on their characters. To do that, we
+    // will consider the sort lists of characters (see `CharSet#characters()`)
+    // of the two sets respectively. We will then lexicographically compare
+    // these lists of characters.
+    // Obviously, we don't actually look at the full list of characters.
+    // CharSets are represented as ranges, and we will take advantage of that.
+    // In lexicographical sorting, we just have to find the first character
+    // that differs in the two sequences, and that's quite simple to do in the
+    // range representation. Further, if one sequence ends before that
+    // character was found, we compare the length of the two sequences. That is
+    // trivial to do in the range form as well.
+
+    const aRanges = a.ranges
+    const bRanges = b.ranges
+    for (let i = 0; i < aRanges.length && i < bRanges.length; i++) {
+        const aR = aRanges[i]
+        const bR = bRanges[i]
+
+        if (aR.min !== bR.min) return aR.min - bR.min
+        if (aR.max !== bR.max) {
+            if (aR.max < bR.max) {
+                // [aR.min .. aR.max]           [...?]
+                // [bR.min .. aR.max .. bR.max]
+
+                // If there is another range for a, then a is larger than b
+                return i + 1 < aRanges.length ? +1 : -1
+
+                // eslint-disable-next-line no-else-return -- x
+            } else {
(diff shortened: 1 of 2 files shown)


ota-meshi:
Thank you for the detailed explanation.

I think it would be best if we could sort it in lexicographically as you say.

> The only problem with this sorting method is that it doesn't work. The "lexicographically smallest string" is only defined for finite sets of strings. I.e. What's the lexicographically smallest string for a+b? ab > aab > aaab > aaaaaaaaaaaab > ... I'm not sure how this could be fixed.

What do you think about limiting the maximum number of repeats? (For example, 10 repeats)
I think it works for most of the regexes that are actually used.

RunDevelopment:
> What do you think about limiting the maximum number of repeats?

That's difficult to do right. E.g. how do we limit the repeats in `((((a+)+)+)+)+`?

---

I was thinking about implementing this a little more. To find the lexicographically smallest string (LSS) in either `a` or `b` but not both, we would need to compute the (symmetric) difference of `a` and `b`. This is a `O(2^n)` operation with regexes and might be a little too slow for us. 

So how about we use a hybrid method? I was thinking about this:
1. Find the LSS for both `a` and `b` respectively. Let's call those strings `a'` and `b'`.
2. If both `a'` and `b'` are defined/were computed successfully:
    1. If `a' < b'`, then return `a < b`. 
    2. If `a' > b'`, then return `a > b`.
3. Otherwise, compare using prefixes as implemented in this PR.

This has a bunch of nice properties:
- If the LSS is defined, it's used in the comparison.
- We don't have to compute any differences, so it's fast.
- The LSS is computed per alternative, so it's easy to cache.
- Two alternatives having the same LSS should be pretty rare in the wild, so we will likely not fallback to prefix-based comparisons too often.

However, there's a problem again: the comparison function isn't transitive. The problem is that based on LSS comparison, an alternative without an LSS is "equal" to all other alternatives. This can be fixed using a little trick: if the LSS of an alternative isn't defined, we compute the prefix and pick the LSS of the prefix. With this little trick, the comparison function becomes transitive.

I'll try implementing this.

ota-meshi:
> That's difficult to do right. E.g. how do we limit the repeats in ((((a+)+)+)+)+?

You are right. The maximum is not a good idea 😓.

The hybrid method sounds good to me :+1:

RunDevelopment:
Alright, I implemented it. 

Computing the LSS is fairly simple for NFAs, so I convert alternatives into NFAs using refa, and then compute the LSS. Since this is fairly costly, I also implemented a "fast" path when the alternative is just a simple sequence of characters.

I tested this on Prism and the results are pretty good. They are mostly (~99%) the same as before, but it finally gets difficult cases like `(?:sample|(?:un)?stable)` right.

ota-meshi (approved):
Amazing! Thank you

PR #452, opened 2022-07-26
Add option for capturing groups to `regexp/optimal-quantifier-concatenation`

### LINKED ISSUE #451: optimal-quantifier-concatenation: False positive for /^(a*).*/u
**Information:**
- ESLint version: 8.20.0
- `eslint-plugin-regexp` version: 1.7.0

**Description**

I think this rule should report `/^a*.*/u`, but not `/^(a*).*/u`.
If I replace `/^(a*).*/u` with `/^.*/u` I will not be able to use capture group.
At least there can be an option to suppress such warnings.

RunDevelopment:
Resolves #451.

As suggested by @ilyub, I added an option to turn off reports related to capturing groups. These reports cannot be fixed in general and might be false positives. Since they are not only false positives and might reveal bugs, capturing groups are still reported by default.

### DIFF
--- a/docs/rules/optimal-quantifier-concatenation.md
+++ b/docs/rules/optimal-quantifier-concatenation.md
@@ -43,7 +43,37 @@ var foo = /\w+(?:(a)|b)*/;
 
 ## :wrench: Options
 
-Nothing.
+
+```json5
+{
+  "regexp/optimal-quantifier-concatenation": [
+    "error",
+    {
+        "capturingGroups": "report"
+    }
+  ]
+}
+```
+
+### `capturingGroups`
+
+The type of concatenation this rule reports might be intentional around capturing groups. This option allows you turn of false unfixable reports around capturing groups.
+
+- `capturingGroups: "report"` (_default_)
+
+  Concatenations around quantifiers will be reported.
+
+- `capturingGroups: "ignore"`
+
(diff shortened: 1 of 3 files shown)


docs/rules/optimal-quantifier-concatenation.md
```diff
@@ -43,7 +43,36 @@ var foo = /\w+(?:(a)|b)*/;
...
+
+### `capturingGroups`
+
+The type of concatenation this rule reports might be intentional around capturing groups. This option allows you turn of false unfixable reports around capturing groups.
+
+- `capturingGroups: "report"` (_default_)
+
+  Concatenations around quantifiers will be reported.
+
+- `capturingGroups: "ignore"`
```
ota-meshi, line 65: The `"ignore"` option also suppresses the next RegExp warning. Is this intentional?

```js
/\d+(\d*)/u
```

It doesn't make sense because the capture group always returns blanks.

RunDevelopment: > It doesn't make sense because the capture group always returns blanks.

Yes, but finding nonsensical capturing groups is not the purpose of this rule. This rule just also finds some bad capturing groups from time to time.

We should add a new rule for finding bad capturing groups.

ota-meshi (approved):
LGTM! Thank you!

Issue #460, opened 2022-08-21
Incorrect `no-super-linear-move` report with `$` anchor


ericcornelissen:
**Information:**
- ESLint version: v8.22.0
- `eslint-plugin-regexp` version: v1.8.0

**Description**

The `regexp/no-super-linear-move` reports the error:

```js
   /[\t\n ]+$/
//  ~~~~~~~~
//  ^ Any attack string /[\t\n ]+/ plus some rejecting suffix will cause
//    quadratic runtime because of this quantifier
```

even though, if I'm not mistaken, the `$` anchor ensures no "rejecting suffix" is possible.

This should be reproducible by placing the above regular expression in any project with the following configuration:

```yaml
# .eslintrc.yml

env:
  browser: false
  es2021: true
plugins:
  - regexp
parserOptions:
  ecmaVersion: latest
  sourceType: module
rules:
  regexp/no-super-linear-move:
    - error
    - report: potential
      ignoreSticky: false
```

Several examples of this bug can also be found at https://github.com/ericcornelissen/shescape/pull/373/files#diff-674954bbcb85fa998a1ef01768e3925bd1a9bbbed52533061bb51f4efdbebbbe.

RunDevelopment:
No, this is correct. The regex `/[\t\n ]+$/` will reject any string `" ".repeat(n) + "a"` in O(n^2) steps. You can read more on this [here](https://codeql.github.com/codeql-query-help/javascript/js-polynomial-redos/#example).

> if I'm not mistaken, the `$` anchor ensures no "rejecting suffix" is possible.

"Rejecting suffix" means some string that gets rejected by both the quantifier and the pattern after the quantifier. Since `$` rejects everything, we can use any character rejected by the quantifier.

ericcornelissen:
Thanks for explaining @RunDevelopment - makes sense now. The [link](https://codeql.github.com/codeql-query-help/javascript/js-polynomial-redos/#example) really helped me understand.

I just now noticed the [Further reading](https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-super-linear-move.html#further-reading) section on the rule's documentation page. Perhaps it could be worth adding the link there?

PR #461, opened 2022-08-21
Add new further reading link to `no-super-linear-move` docs

RunDevelopment:
As suggested by @ericcornelissen [here](https://github.com/ota-meshi/eslint-plugin-regexp/issues/460#issuecomment-1221559414).


ota-meshi (approved):
Thank you!

PR #462, opened 2022-08-21
Add `minAlternatives` option for `prefer-character-class`

### LINKED ISSUE #459: Report 2 single-character alternative groups with regexp/prefer-character-class
**Description**

I noticed that the `regexp/prefer-character-class` rule does not report on groups with 2 single-character alternatives, e.g. `(?:a|b)`. I would expect it do this, I didn't spot any justification for not reporting these in the docs (if there is, please let me know :slightly_smiling_face:)

Alternatively (and/or to avoid breaking compatibility), a solution could be to add a numeric option to the rule to configure the minimum number of single-character alternatives for which the rule should report. The default for this option would then probably be `3` (same as the current behavior), and could be configured to `2` (or less) if desired.

RunDevelopment:
Resolves #459.

### DIFF
--- a/docs/rules/prefer-character-class.md
+++ b/docs/rules/prefer-character-class.md
@@ -44,7 +44,22 @@ var foo = /(\w|\d)+:/
 
 ## :wrench: Options
 
-Nothing.
+```json5
+{
+  "regexp/prefer-character-class": [
+    "error",
+    {
+        "minAlternatives": 3
+    }
+  ]
+}
+```
+
+### `minAlternatives: integer`
+
+This number controls how many character alternatives have to be present for them to be merged. By default, there need to at least 3 alternatives.
+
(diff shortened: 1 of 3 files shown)


docs/rules/prefer-character-class.md
```diff
@@ -44,7 +44,22 @@ var foo = /(\w|\d)+:/
...
+    {
+        "minAlternatives": 3
+    }
+  ]
+}
+```
+
+### `minAlternatives: integer`
+
+This number controls how many character alternatives have to be present for them to be merged. By default, there need to at least 3 alternatives.
```
ericcornelissen, line 60: Small typo:

```suggestion
This number controls how many character alternatives have to be present for them to be merged. By default, there need to be at least 3 alternatives.
```

RunDevelopment: Thank you!

ota-meshi (approved):
LGTM! Thank you!

PR #473, opened 2022-11-03
chore: use lockfile v2

ota-meshi:
This PR changes the lockfile version to v2. Probably node v14 is rarely used, so using v2 is fine.
There are also some fixes associated with package upgrades.


tests/lib/rules/no-useless-flag.ts
```diff
@@ -872,9 +874,11 @@ describe("Don't conflict even if using the rules together.", () => {
             code: String.raw`/[\s\S]*/s;`,
             output: String.raw`/[\s\S]*/;`,
             rulesConfig: {
+                /* eslint-disable @typescript-eslint/naming-convention -- rule names */
```
RunDevelopment, line 877: Most changes are related to this rule. Was this rule added to a recommended config or why does it cause so many errors now? It seems a little annoying...

Since this is a purely stylistic rule, we may even consider disabling the rule entirely.

ota-meshi: Thank you for checking this PR!
I will change `@typescript-eslint/naming-convention` to turn off.

RunDevelopment (approved):

PR #474, opened 2022-11-10
Add `regexp/no-missing-g-flag` rule

### LINKED ISSUE #472: Rule proposal: `no-missing-g-flag`
**Motivation**

When calling [`String#replaceAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) and [`String#matchAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) with non-global RegExp, it will be a runtime error. I hope to catch this pattern during linting.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#exceptions
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#exceptions

**Description**

Forbid calling `String#replaceAll()` or `String#matchAll()` with non-global pattern.

**Examples**

```js
/* ✓ GOOD */
foo.replaceAll(" ", '')
foo.replaceAll(/\s/g, '')
foo.matchAll(/\s/g)

/* ✗ BAD */
foo.replaceAll(/\s/, '')
foo.matchAll(/\s/)
```

ota-meshi:
This PR adds `regexp/no-missing-g-flag` rule.

close #472

### DIFF
--- a/lib/rules/no-missing-g-flag.ts
+++ b/lib/rules/no-missing-g-flag.ts
@@ -1,5 +1,4 @@
-import type { RegExpVisitor } from "regexpp/visitor"
-import type { RegExpContext } from "../utils"
+import type { RegExpContext, UnparsableRegExpContext } from "../utils"
 import { createRule, defineRegexpVisitor } from "../utils"
 import type { ExpressionReference } from "../utils/ast-utils"
 import {
@@ -60,16 +59,16 @@ export default createRule("no-missing-g-flag", {
         const { strictTypes } = parseOption(context.options[0])
         const typeTracer = createTypeTracker(context)
 
-        /**
-         * Create visitor
-         */
-        function createVisitor(
-            regexpContext: RegExpContext,
-        ): RegExpVisitor.Handlers {
-            const { regexpNode, flags, ownsFlags } = regexpContext
+        /** The logic of this rule */
+        function visit(regexpContext: RegExpContext | UnparsableRegExpContext) {
+            const { regexpNode, flags, flagsString } = regexpContext
 
-            if (flags.global || !ownsFlags) {
-                return {}
+            if (
+                flags.global ||
+                // We were unable to determine which flags were used.
+                flagsString == null
+            ) {
+                return
             }
 
             for (const ref of extractExpressionReferences(
@@ -78,7 +77,6 @@ export default createRule("no-missing-g-flag", {
             )) {
                 verifyExpressionReference(ref, regexpContext)
             }
-            return {}
         }
 
         /**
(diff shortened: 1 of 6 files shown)


lib/rules/no-missing-g-flag.ts
```diff
@@ -0,0 +1,138 @@
...
+
+        /**
+         * Create visitor
+         */
+        function createVisitor(
+            regexpContext: RegExpContext,
+        ): RegExpVisitor.Handlers {
+            const { regexpNode, flags, ownsFlags } = regexpContext
+
+            if (flags.global || !ownsFlags) {
```
RunDevelopment, line 71: Should we really require `ownsFlags`? If a pattern doesn't own its flags, it will still cause a runtime error. The only difference is that we can't provide a fix (and `fixReplaceFlags` will automatically handle this).

lib/rules/no-missing-g-flag.ts
```diff
@@ -0,0 +1,138 @@
...
+                if (node.arguments[0] !== regexpNode) {
+                    // It can only be safely auto-fixed if it is defined directly in the argument.
+                    return null
+                }
+                return fixReplaceFlags(`${flagsString || ""}g`, false)
+            }
+        }
+
+        return defineRegexpVisitor(context, {
+            createVisitor,
```
RunDevelopment, line 135: Since don't actually look at the pattern, we don't have to require a valid pattern at all. I think this rule should support unknown and invalid patterns as well. `sort-flags` also works this way.

RunDevelopment (changes requested):
Looking good! Just 2 comments.

lib/rules/no-missing-g-flag.ts
ota-meshi: Thank you for checking this out.
I should have used `flagsString == null` instead.

lib/rules/no-missing-g-flag.ts
ota-meshi: I forgot I had that option 😓. I would change this PR to also use visitInvalid and visitUnknown.

RunDevelopment (approved):
Looks good! Thank you!

lib/rules/no-missing-g-flag.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export default createRule("no-missing-g-flag", {
+    meta: {
+        docs: {
+            description:
+                "disallow missing `g` flag in patterns used in `String#matchAll` and `String#replaceAll`",
+            category: "Possible Errors",
+            // TODO Switch to recommended in the major version.
+            // recommended: true,
```
fisker, line 39: Why this not recommend?

RunDevelopment: Because that would be a breaking change. Existing users that update the plugin might suddenly find their CI failing if we included it.

fisker: Understood, thanks.

PR #480, opened 2022-11-19
Improve `prefer-lookaround` rule to report when there are leading/trailing assertions

### LINKED ISSUE #478: `prefer-lookaround` should ignore `^` and `$`
**Information:**
- ESLint version: 8.27.0
- `eslint-plugin-regexp` version: 7.31.10

**Description**

```js
var str = 'JavaScript'.replace(/Java(Script)/g, 'Type$1')
var str = 'JavaScript'.replace(/Java(Script$)/g, 'Type$1')
var str = 'JavaScript'.replace(/Java(Script)$/g, 'Type$1')
```

The first two are reported, but the 3rd is not.

I think 

```js
var str = 'JavaScript'.replace(/Java(?=Script)$/g, 'Type')
```

should result the same. Will adding support for this case break things?

ota-meshi:
close  #478

### DIFF
--- a/lib/rules/prefer-lookaround.ts
+++ b/lib/rules/prefer-lookaround.ts
@@ -258,7 +258,7 @@ function isTrailingAssertion(element: Element): element is TrailingAssertion {
 }
 
 /** Checks whether the given element is simple (single alternative, and positive) lookaround assertion or not. */
-function isSingleLookaroundAssertion(
+function isSimpleLookaroundAssertion(
     element: Element,
 ): element is LookaroundAssertion & {
     negate: false
@@ -330,7 +330,7 @@ function parsePatternElements(node: Pattern): {
                 !leadingAssertion
                     ? ""
                     : // If the leading assertion is simple lookbehind assertion, unwrap the parens.
-                    isSingleLookaroundAssertion(leadingAssertion)
+                    isSimpleLookaroundAssertion(leadingAssertion)
                     ? leadingAssertion.alternatives[0].raw
                     : leadingAssertion.raw
             }${capturingGroup.alternatives.map((a) => a.raw).join("|")})`,
@@ -364,7 +364,7 @@ function parsePatternElements(node: Pattern): {
                 !trailingAssertion
                     ? ""
                     : // If the trailing assertion is simple lookahead assertion, unwrap the parens.
-                    isSingleLookaroundAssertion(trailingAssertion)
+                    isSimpleLookaroundAssertion(trailingAssertion)
                     ? trailingAssertion.alternatives[0].raw
                     : trailingAssertion.raw
             })`,
--- a/lib/rules/prefer-lookaround.ts
+++ b/lib/rules/prefer-lookaround.ts
@@ -1,5 +1,16 @@
 import type { RegExpVisitor } from "regexpp/visitor"
-import type { CapturingGroup, Element } from "regexpp/ast"
+import type {
+    Alternative,
+    CapturingGroup,
+    EdgeAssertion,
+    Element,
+    Group,
+    LookaheadAssertion,
+    LookaroundAssertion,
+    LookbehindAssertion,
+    Pattern,
+    WordBoundaryAssertion,
+} from "regexpp/ast"
 import type { RegExpContext } from "../utils"
 import { createRule, defineRegexpVisitor } from "../utils"
 import { createTypeTracker } from "../utils/type-tracker"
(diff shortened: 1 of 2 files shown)


lib/rules/prefer-lookaround.ts
```diff
@@ -172,6 +185,205 @@ function getSideEffectsWhenReplacingCapturingGroup(
...
+        )
+    } else if (element.type === "Group") {
+        if (element.alternatives.length === 0) return false
+        for (const alt of element.alternatives) {
+            if (alt.elements.length !== 1) return false
+            if (!isLeadingAssertion(alt.elements[0])) return false
+        }
+        return true
+    }
+    return false
```
RunDevelopment, line 237: Why are there these requirements for leading and trailing assertions? As far as I see it, any zero-length element will work. E.g. `/foo(bar)(?:^|$|a{0}\b|(?=x)|(?<=y))/` -> `/foo(?=bar(?:^|$|a{0}\b)|(?=x)|(?<=y))/` is a valid transformation. Of course, the regex doesn't make sense, but we have other rules that will report these problems.

lib/rules/prefer-lookaround.ts
```diff
@@ -172,6 +185,205 @@ function getSideEffectsWhenReplacingCapturingGroup(
...
+    }
+
+    return {
+        elements,
+        start,
+        end,
+    }
+
+    /** Convert leading/trailing assertion to lookaround assertion pattern text. */
+    function assertionToLookaroundAssertionPatternText(
```
RunDevelopment, line 371: This function has to make sure that the lookaheads/lookbehinds it unwraps look into the same direction as their parent. E.g. `foo(bar)(?<=z\w+)` -> `foo(barz\w+)` is wrong, it has to be `foo(bar(?<=z\w+))`.

Also, should this rule even unwrap lookarounds? Transforming `(?=foo(?=bar))` -> `(?=foobar)` seems like something a dedicated rule should do. Maybe [`no-trivially-nested-assertion`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-trivially-nested-assertion.html) could do it?

RunDevelopment (commented):
Thank you for implementing this!

I find 2 things we should talk about.

lib/rules/prefer-lookaround.ts
RunDevelopment: Oh, I didn't see #481. That also works for me :)

lib/rules/prefer-lookaround.ts
ota-meshi: Thank you for your review! 
I didn't come up with an idea to check for zero-length element.
I will change to use that method.

lib/rules/prefer-lookaround.ts
ota-meshi: > Also, should this rule even unwrap lookarounds?

In my opinion, we need to unwrap the simplest lookarounds. prefer-lookaround rule generate lookaround assertions with autofix.
I think the new report after autofix should combine the assertions made by the rule and autofix them.

```js
var str = 'JavaScript'.replace(/Java(S)(c)(r)(i)(p)(t)/g, 'Type$1$2$3$4$5$6')
// ↓
var str = 'JavaScript'.replace(/Java(S)(c)(r)(i)(p)(?=t)/g, 'Type$1$2$3$4$5')
// ↓
var str = 'JavaScript'.replace(/Java(S)(c)(r)(i)(?=pt)/g, 'Type$1$2$3$4')
// ↓
var str = 'JavaScript'.replace(/Java(S)(c)(r)(?>=ipt)/g, 'Type$1$2$3')
// ↓
// ...
// ↓
var str = 'JavaScript'.replace(/Java(?>=Script)/g, 'Type')
```

RunDevelopment: Alright, then let's unwrap simple lookarounds.

lib/rules/prefer-lookaround.ts
```diff
@@ -172,6 +180,190 @@ function getSideEffectsWhenReplacingCapturingGroup(
     }
 }
 
+/** Checks if the given element is a zero length element that can be placed as a leading/trailing element. */
+function isLeadingTrailingElement(element: Element): boolean {
+    if (element.type === "CapturingGroup") return false
```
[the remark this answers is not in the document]
ota-meshi: You're right! I will change this PR.

RunDevelopment (approved):
Thank you @ota-meshi!

PR #482, opened 2022-11-20
Add `regexp/no-extra-lookaround-assertions` rule

### LINKED ISSUE #481: New rule: `no-useless-lookaround-assertions`
**Motivation**

We can define unnecessary lookaround assertions within lookaround assertions, so I would like a rule to report this and fix it automatically.

```js
'JavaScript'.replace(/Java(?=Scrip(?=t))/u, 'Type');
'JavaScript'.replace(/(?<=(?<=J)ava)Script/u, '');
```

**Description**

New rule report unnecessary lookaround assertions within lookaround assertions.

**Examples**

```js
/* ✗ BAD */
'JavaScript'.replace(/Java(?=Scrip(?=t))/u, 'Type');
'JavaScript'.replace(/(?<=(?<=J)ava)Script/u, '');

/* ✓ GOOD */
'JavaScript'.replace(/Java(?=Script)/u, 'Type');
'JavaScript'.replace(/(?<=Java)Script/u, '');
```

ota-meshi:
close #481


fisker:
How about `regexp/no-useless-assertions`?

ota-meshi:
The description of no-useless-assertions says:

> Some assertion are unnecessary because the rest of the pattern forces them to always be accept (or reject).

The new rule don't report always accept/reject patterns, so I think the rules have different goals.

lib/rules/no-useless-lookaround-assertions.ts
```diff
@@ -0,0 +1,115 @@
...
+                loc: getRegexpLocation(assertion),
+                messageId: "unexpected",
+                data: {
+                    kind: assertion.kind,
+                },
+                fix: fixReplaceNode(assertion, () => {
+                    const alternatives = assertion.alternatives.map(
+                        (alt) => alt.raw,
+                    )
+                    if (alternatives.length <= 1) {
```
RunDevelopment, line 101: Please make this more explicit.

```suggestion
                    if (alternatives.length === 1) {
```

lib/rules/no-useless-lookaround-assertions.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        /** Verify for lookbehind assertion */
+        function verifyForLookbehind(
+            regexpContext: RegExpContext,
+            assertion: LookbehindAssertion,
+        ) {
+            for (const alternative of assertion.alternatives) {
+                const first = alternative.elements[0]
+                if (
+                    first?.type !== "Assertion" ||
```
RunDevelopment, line 73: While this does work correctly for empty alternatives, I don't like that TypeScript thinks that the type is `first: Element`. If I removed the `?`, TypeScript wouldn't be able to catch the bug.

The [`at` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) is what would be ideal here, I think. But I don't think that we can use it yet because of browser/node support. Maybe we could make our own `at` method and use that?

lib/rules/no-useless-lookaround-assertions.ts
```diff
@@ -0,0 +1,115 @@
...
+                        verifyForLookahead(regexpContext, aNode)
+                    } else if (aNode.kind === "lookbehind") {
+                        verifyForLookbehind(regexpContext, aNode)
+                    }
+                },
+            }
+        }
+
+        /** Verify for lookahead assertion */
+        function verifyForLookahead(
```
RunDevelopment, line 45: Is it really necessary to have 2 methods for verifying? They only differ in which index they access and what `.kind` they require. A combined verify method could look like this:

```ts
        /** Verify for lookaround assertion */
        function verify(
            regexpContext: RegExpContext,
            assertion: LookaroundAssertion,
        ) {
            for (const alternative of assertion.alternatives) {
                const nested = alternative.elements.at(
                    assertion.kind === "lookahead" ? -1 : 0,
                )
                if (
                    nested?.type === "Assertion" &&
                    nested.kind === assertion.kind &&
                    !nested.negate
                ) {
                    reportLookaroundAssertion(regexpContext, nested)
                }
            }
        }
```

RunDevelopment (changes requested):

ota-meshi:
@RunDevelopment Thank you for the your review!
I made changes to this PR. Could you please check again?

RunDevelopment:
Thank you @ota-meshi.

I think the rule is implemented correctly, but I think the name/message might be a little confusing. The name "no useless lookaround assertions" and the message "This assertion is useless." might make it seem like reported assertions can be removed. E.g. `(?=foo(?=bar))` -> `(?=foo)`. Obviously, this is not what we mean, but I worry that some users might understand that way.

Maybe we could change the message to something like "This assertion is useless and can be inlined" or "This assertion is useless and can be converted into a group"? If we go with those messages, then we should probably also update the docs to use this language too. What do you think?

ota-meshi:
I think the message you suggested is good! I will change it to use that message.
Do you have other good ideas for rule name?

What do you think about naming it no-extra-lookaround-assertions, in reference to no-extra-parens?
https://eslint.org/docs/latest/rules/no-extra-parens

RunDevelopment:
> What do you think about naming it no-extra-lookaround-assertions, in reference to no-extra-parens?

That is a really good idea! Let's go with that.

ota-meshi:
I have changed this PR. Could you please check again?

RunDevelopment (approved):
Thank you @ota-meshi!

PR #483, opened 2022-11-20
Add `no-misleading-capturing-group` rule and improve `optimal-quantifier-concatenation`

### LINKED ISSUE #453: New rule to find bad capturing groups
**Description**
Add a new rule to find bad (= obviously or likely incorrect) capturing groups. 

Bad capturing groups are:
- Capturing groups that do not behave as their pattern would suggest. E.g. `/\d+(\d*)/` will always capture the empty string despite containing the pattern `\d*`. In general, we should verify that the language of all possible captured strings is equal to the language of the pattern of a capturing group.
- Capturing groups that likely behave incorrectly. E.g. `/^(a*).+/u` likely behaves incorrectly for the input string `a`. In general, we should verify that a capturing group cannot exchange characters with the pattern after it. 
   This check should be behind an option to turn it off. It will find **likely** bad capturing groups, so false positives are to be expected.

Reference: #451, https://github.com/ota-meshi/eslint-plugin-regexp/pull/452/files#r930616165

**Examples**

```js
/* ✗ BAD */
var foo = /^(a*).+/u
var foo = /\d+(\d*)/u
```

RunDevelopment:
Resolves #453.

Regarding the new simplify quantifier method: The implementation is quite complex. It seems to work, but I still have to test it on Prism's regexes.

This PR is still a work in progress.


ota-meshi (approved):
Thank you for the great work!

PR #485, opened 2022-11-21
Chore: use vitepress instead of vuepress as site builder

ota-meshi:
This PR will replace the site builder from vuepress to vitepress.
I think we have to do it because vuepress can't build without the flag on Node18.

In order to build with vitepress, there are places where the code of the plugin itself has been refactored.

There are some advantages to changing to using vuepress.
1. Reduced startup time in development mode.
2. No longer need to build the plugin before `npm run docs:watch`.


RunDevelopment (approved):
I have no experience with Vue, so I'm just going to trust that everything is correct :)

Just one questions: Are all URLs going to stay the same? I don't want us to break any old links.

ota-meshi:
> Are all URLs going to stay the same?

Yes, the site URLs remain the same. (README.md replaced by index.md for that)

PR #486, opened 2022-11-21
Fixed false negatives around edges and added second phase of reporting

### LINKED ISSUE #479: `no-useless-assertions` false negative around `$`
#478 included the regex: `/Java(?=Script)$/`. This regex will always reject because the 2 assertions conflict each other. `no-useless-assertions` is supposed to report this but doesn't. The rule does find the problem in ``/Java$(?=Script)/`` though.

image
image

RunDevelopment:
Fixes #479.

This PR includes 2 changes:
1. I improved the handling of edges. Previously, all verify function would just immediately return without reporting anything if the next character included the start/end (edge) of the input string. I likely did this because the logic around string edges is a bit more complex.
2. In order to fully report all useless assertion in the examples [here](https://github.com/ota-meshi/eslint-plugin-regexp/issues/479#issuecomment-1321446605), I had to add assertion reordering. E.g. to detect that the `\b` in `/\b^a/` is useless, `\b` has to know that there is a `^` after it. The solution here is to (conceptually) reorder the assertions to `/^\ba/` which allows `\b` to "see" `^`.
    However, assertion reordering also causes overreporting. E.g. Since both `$`s in `/$$/` can see each other, both will be reported even though only one is actually useless. To fix this, reporting is done in 2 phases. In the first phase, we do everything just like we did before. In the second phase, we reorder while being careful not to reorder already reported assertions.

### DIFF
--- a/lib/rules/no-useless-assertions.ts
+++ b/lib/rules/no-useless-assertions.ts
@@ -2,11 +2,18 @@ import type { RegExpVisitor } from "regexpp/visitor"
 import type {
     Assertion,
     EdgeAssertion,
+    Element,
     LookaroundAssertion,
     WordBoundaryAssertion,
 } from "regexpp/ast"
+import { visitRegExpAST } from "regexpp"
 import type { RegExpContext } from "../utils"
 import { createRule, defineRegexpVisitor } from "../utils"
+import type {
+    MatchingDirection,
+    ReadonlyFlags,
+    FirstLookChar,
+} from "regexp-ast-analysis"
 import {
     Chars,
     getFirstCharAfter,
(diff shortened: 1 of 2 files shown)


lib/rules/no-useless-assertions.ts
```diff
@@ -273,24 +397,60 @@ export default createRule("no-useless-assertions", {
...
+                        onAssertionEnter(assertion) {
+                            if (reported.has(assertion)) {
+                                return
+                            }
+                            verifyAssertion(
+                                assertion,
+                                reorderingGetFirstCharAfter,
+                            )
+                        },
+                    })
```
ota-meshi, line 453: By collecting the nodes visited in the traverse, I think it will be unnecessary to trace from the root in the second traverse.


```ts

            const allAssertions: Assertion[] = []
            return {
                onAssertionEnter(assertion) {
                    // Phase 1:
                    // ...
                    verifyAssertion(assertion, getFirstCharAfter)
                    allAssertions.push(assertion)
                },
                onPatternLeave() {
                    // Phase 2:
                    // ...
                    const reorderingGetFirstCharAfter =
                        createReorderingGetFirstCharAfter(reported)
                    for (const assertion of allAssertions) {
                        if (reported.has(assertion)) {
                            continue
                        }
                        verifyAssertion(assertion, reorderingGetFirstCharAfter)
                    }
                },
            }
```

ota-meshi (changes requested):

lib/rules/no-useless-assertions.ts
RunDevelopment: Good point!

PR #493, opened 2022-11-26
Automate docs with eslint-doc-generator

### LINKED ISSUE #476: docs: use `eslint-doc-generator`
@bmish created this awesome library that brings a unified way of rule docs
https://github.com/bmish/eslint-doc-generator

bmish:
Fixes #476.

[eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) is a CLI tool I built for automating the generation of the README rules list table and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of custom rules through better documentation and streamline the process of adding new rules). It has 100% test coverage and is [used](https://github.com/bmish/eslint-doc-generator#users) in many of the top ESLint plugins already. It replaces the existing README rules list / rule doc header documentation generation scripts in this repository.

Notable changes:
* Moved `meta.docs.replacedBy` to the [standard ESLint rule property](https://eslint.org/docs/latest/developer-guide/working-with-rules) `meta.replacedBy` for indicating deprecated rule replacement rules
* More clearly indicate:
   * Whether a rule is deprecated
   * Whether a rule provides suggestions
   * Whether a rule *warns* in the recommended config as opposed to *errors*
* The `Version` section in the rule doc footer now indicates when a rule has not been released yet. This is because the scripts in this repo no longer manage the rule doc header. I think this is more organized and consistent anyway.


docs/rules/index.md
```diff
@@ -4,108 +4,106 @@ sidebarDepth: 0
...
+| [prefer-unicode-codepoint-escapes](https://ota-meshi.github.io/eslint-plugin-regexp/rules/prefer-unicode-codepoint-escapes.html) | enforce use of unicode codepoint escapes                               | ✅  |    | 🔧 |    |    |
+| [prefer-w](https://ota-meshi.github.io/eslint-plugin-regexp/rules/prefer-w.html)                                                 | enforce using `\w`                                                     | ✅  |    | 🔧 |    |    |
+| [sort-character-class-elements](https://ota-meshi.github.io/eslint-plugin-regexp/rules/sort-character-class-elements.html)       | enforces elements order in character class                             |    |    | 🔧 |    |    |
+| [sort-flags](https://ota-meshi.github.io/eslint-plugin-regexp/rules/sort-flags.html)                                             | require regex flags to be sorted                                       | ✅  |    | 🔧 |    |    |
+| [unicode-escape](https://ota-meshi.github.io/eslint-plugin-regexp/rules/unicode-escape.html)                                     | enforce consistent usage of unicode escape or unicode codepoint escape |    |    | 🔧 |    |    |
 
-### Deprecated
```
ota-meshi, line 100: I want deprecated to be a separate list just like the list in ESLint core. Can we do that?
https://eslint.org/docs/latest/rules/#deprecated

docs/rules/index.md
```diff
@@ -4,108 +4,106 @@ sidebarDepth: 0
...
-| [regexp/no-useless-backreference](./no-useless-backreference.md) | disallow useless backreferences in regular expressions | :star: |
-| [regexp/no-useless-dollar-replacements](./no-useless-dollar-replacements.md) | disallow useless `$` replacements in replacement string | :star: |
-| [regexp/strict](./strict.md) | disallow not strictly valid regular expressions | :star::wrench: |
+| Name                                                                                                                                       | Description                                                                                | 💼 | ⚠️ | 🔧 | 💡 | ❌  |
+| :----------------------------------------------------------------------------------------------------------------------------------------- | :----------------------------------------------------------------------------------------- | :- | :- | :- | :- | :- |
+| [confusing-quantifier](https://ota-meshi.github.io/eslint-plugin-regexp/rules/confusing-quantifier.html)                                   | disallow confusing quantifiers                                                             |    | ✅  |    |    |    |
```
ota-meshi, line 20: Is it possible to format the URL to be the original relative path + *.md?
This is because the SPA routing by vitepress will not work properly if we specify the URL.

```suggestion
| [confusing-quantifier](./no-contradiction-with-assertion.md)                                   | disallow confusing quantifiers                                                             |    | ✅  |    |    |    |
```

README.md
```diff
@@ -97,116 +97,109 @@ The `plugin:regexp/all` config enables all rules. It's meant for testing, not fo
...
 ## :white_check_mark: Rules
 
-<!--RULES_SECTION_START-->
+<!-- begin auto-generated rules list -->
 
-The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) automatically fixes problems reported by rules which have a wrench :wrench: below.
-The rules with the following star :star: are included in the `plugin:regexp/recommended` config.
+💼 Configurations enabled in.\
+⚠️ Configurations set to warn in.\
+✅ Set in the `recommended` configuration.\
```
ota-meshi, line 104: I would like to clarify the name of the configuration in the list and in each rule document.

```suggestion
✅ Set in the `plugin:regexp/recommended` configuration.\
```

It seems that the eslint core clearly states that it is included in `"eslint:recommended"`.

https://eslint.org/docs/latest/rules/constructor-super

ota-meshi (changes requested):
Thank you for this PR! I have some requests.

docs/rules/index.md
```diff
@@ -4,108 +4,106 @@ sidebarDepth: 0
...
-<!-- This file is automatically generated in tools/update-docs-rules-index.js, do not change! -->
+💼 Configurations enabled in.\
+⚠️ Configurations set to warn in.\
+✅ Set in the `recommended` configuration.\
+🔧 Automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/user-guide/command-line-interface#--fix).\
+💡 Manually fixable by [editor suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).\
+❌ Deprecated.
 
-### Possible Errors
+## Best Practices
```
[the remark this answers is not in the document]
bmish: We're using `--split-by meta.docs.category` [option](https://github.com/bmish/eslint-doc-generator#configuration-options) to achieve this. The sub-lists are ordered alphabetically. I'm not sure how we could achieve the original ordering using a generic option (let me know if you have ideas).

docs/rules/index.md
bmish: I just filed https://github.com/bmish/eslint-doc-generator/issues/291 which will be necessary to achieve that. We can only use `--split-by` to split the rules list by one property currently, and we're already using that for `--split-by meta.docs.category`. Note that eslint-plugin-jest has an example with a [rules list](https://github.com/jest-community/eslint-plugin-jest#rules) where we use `--split-by meta.deprecated`.

README.md
bmish: I filed https://github.com/bmish/eslint-doc-generator/issues/292 to consider this as an option. See my reasoning in the issue for why I don't think it should be the default behavior. Let me know if you still want this option to be implemented.

docs/rules/prefer-t.md
```diff
@@ -7,10 +7,13 @@ since: "v0.1.0"
 ---
 # regexp/prefer-t
 
-> enforce using `\t`
+❌ This rule is deprecated. It was replaced by [`control-character-escape`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/control-character-escape.html).
```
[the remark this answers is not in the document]
bmish: I agree with this, filed https://github.com/bmish/eslint-doc-generator/issues/293. We'll need to include the prefix in the rule names in `meta.replacedBy` for it to work.

docs/rules/index.md
bmish: The requirements for this are a bit confusing. Is the full rule URL only supposed to be used in README.md? I switched to using the full URL everywhere (using the `urlRuleDoc` option) so that users are sent to the external site regardless of what GitHub page they start from.

These are the places where we have rule links:
* README.md
* docs/rules/index.md
* docs/rules/{name}.md (some rules have links to their replacement links)

ota-meshi: > Is the full rule URL only supposed to be used in README.md?

Yes. Only README.md uses URLs. This is to link to the documentation page from the package page of npm. `docs/*` are files on the documentation page. They must be linked with relative paths.

If it's difficult, we can consider using a regexp to replace it after processing.

docs/rules/index.md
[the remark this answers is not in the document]
ota-meshi: Hmm, categories are free for plugins, so I think it's hard to decide on a general sort. I think the only way to solve the requirement is to specify the compare function in the config.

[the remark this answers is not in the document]
bmish: Good idea, filed here: https://github.com/bmish/eslint-doc-generator/issues/294

docs/rules/index.md
bmish: I filed https://github.com/bmish/eslint-doc-generator/issues/295 to consider an option for that behavior. Do you think this is a common pattern for plugins that use a static site generator?

ota-meshi: I mostly use vuepress or vitepress so I'm not sure how other static site generators handle links😥

docs/rules/no-empty-alternative.md
```diff
@@ -40,7 +42,7 @@ Nothing.
 
 ## :heart: Compatibility
 
-This rule was taken from [eslint-plugin-clean-regex].  
+This rule was taken from [eslint-plugin-clean-regex].
```
RunDevelopment, line 45: This is a bit of a problem. Ending a line with 2 spaces indicates a line break in markdown, so we can't always trim trailing spaces.

Does eslint-doc-generator do this automatically? If so, this needs to be fixed. I think trimming trailing spaces is good, but it needs to be careful about line breaks.

docs/rules/index.md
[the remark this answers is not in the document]
RunDevelopment: Another option would be to allow users to specify an array of categories (e.g. `categorySort: ["Possible Errors", "Best Paractices", "Stylistic issues"]`). Categories not specified by the user could come in alphabetical order after the specified ones.

This would be a more declarative and JSON-compatible solution.

docs/rules/no-empty-alternative.md
bmish: Interesting. Using trailing spaces for linebreaks seems very fragile as many people including myself have their editors set to automatically trim trailing spaces. So this change came from my editor, not eslint-doc-generator. I can revert this change, but I would suggest using a trailing backslash `\` as a less-fragile and more obvious way to force a linebreak.

README.md
RunDevelopment: I also think that it should be `plugin:regexp/recommended`. Sure, it's longer, but it also makes it easier for users to copy the string and include it in their ESLint config.

bmish: Since you both are in favor of that format, I'm open to adding an option for it. However, one alternate pattern I have seen is that the `## Configs` section in the README lists the configs with more details including how the config would be used (`plugin:regexp/recommended`) so that we don't necessarily need to mention the full details everywhere else in the docs that the config is mentioned.

docs/rules/no-empty-alternative.md
RunDevelopment: Uhhh, I really like the trailing `\` solution, never knew this existed. I actually also have me editor setup to trim trailing spaces and needed to be careful not to commit those changes every time.

Then let's change the `<space><space>` to `<space><backslash>`.

docs/rules/prefer-t.md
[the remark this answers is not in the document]
bmish: This has been resolved.

docs/rules/no-empty-alternative.md
bmish: I don't think there's any need for a space before the backslash. Here's my separate PR to fix this among other things: https://github.com/ota-meshi/eslint-plugin-regexp/pull/494

RunDevelopment: I just wanted a space because I think it would look nicer. But not having it is okay too.

docs/rules/index.md
bmish: I have solved this by defining a `ruleListSplit` function in the config file using the new feature of the [`--rule-list-split` config option](https://github.com/bmish/eslint-doc-generator#configuration-options) from eslint-doc-generator [1.2.0](https://github.com/bmish/eslint-doc-generator/releases/tag/v1.2.0). The new sections should now match the old sections.

docs/rules/index.md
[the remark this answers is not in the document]
bmish: I have solved this by defining a `ruleListSplit` function in the config file using the new feature of the [`--rule-list-split` config option](https://github.com/bmish/eslint-doc-generator#configuration-options) from eslint-doc-generator [1.2.0](https://github.com/bmish/eslint-doc-generator/releases/tag/v1.2.0). The new sections should now match the old sections.

docs/rules/index.md
RunDevelopment: Looks good. Thank you!

README.md
bmish: This has now been addressed using a new option: [`configFormat: 'plugin-colon-prefix-name'`](https://github.com/bmish/eslint-doc-generator#--config-format)

https://github.com/bmish/eslint-doc-generator/releases/tag/v1.3.0

RunDevelopment: Thank you @bmish!

docs/rules/index.md
bmish: I have addressed this by providing a `urlRuleDoc` function in `.eslint-doc-generatorrc.js` to ensure the URL is only used for rule doc links in `README.md` and otherwise the relative path is used. Please take a look.

https://github.com/bmish/eslint-doc-generator/releases/tag/v1.4.0

All feedback should be addressed at this point.

RunDevelopment: Thank you, @bmish!

The relative paths are a bit suboptimal (they are basically `..`-ing out, just to go back into the same directory), but, aesthetics aside, this shouldn't be harmful.

bmish: Yeah I have one bug filed https://github.com/bmish/eslint-doc-generator/issues/284 about resolving the paths better but luckily it still works.

RunDevelopment: Nope, I take that back, they are harmful. The problem is that vitepress (which starts the local server) uses `docs` as the root directory, so we can't go out and back into it.

I'll make a new discussion for this.

docs/rules/index.md
```diff
@@ -4,109 +4,111 @@ sidebarDepth: 0
...
-| [regexp/no-useless-assertions](./no-useless-assertions.md) | disallow assertions that are known to always accept (or reject) | :star: |
-| [regexp/no-useless-backreference](./no-useless-backreference.md) | disallow useless backreferences in regular expressions | :star: |
-| [regexp/no-useless-dollar-replacements](./no-useless-dollar-replacements.md) | disallow useless `$` replacements in replacement string | :star: |
-| [regexp/strict](./strict.md) | disallow not strictly valid regular expressions | :star::wrench: |
+| Name                                                                                             | Description                                                                             | 💼 | ⚠️ | 🔧 | 💡 | ❌  |
+| :----------------------------------------------------------------------------------------------- | :-------------------------------------------------------------------------------------- | :- | :- | :- | :- | :- |
+| [no-contradiction-with-assertion](../../docs/rules/no-contradiction-with-assertion.md)           | disallow elements that contradict assertions                                            |    |    |    | 💡 |    |
```
RunDevelopment, line 20: As mentioned [here](https://github.com/ota-meshi/eslint-plugin-regexp/pull/493#discussion_r1057349245), these relative links are broken. The problem is that vitepress uses `docs/` as the root directory.

To fix this, please optimize/shorten relative paths.

bmish: Good catch, will attempt to fix as part of https://github.com/bmish/eslint-doc-generator/issues/284.

bmish: This should be fixed now: https://github.com/bmish/eslint-doc-generator/releases/tag/v1.4.1

Please verify that the rule links look correct.

RunDevelopment: Links work perfectly now. Thank you!

README.md
```diff
@@ -97,116 +97,114 @@ The `plugin:regexp/all` config enables all rules. It's meant for testing, not fo
...
-The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) automatically fixes problems reported by rules which have a wrench :wrench: below.
-The rules with the following star :star: are included in the `plugin:regexp/recommended` config.
-
-<!--RULES_TABLE_START-->
+💼 Configurations enabled in.\
+⚠️ Configurations set to warn in.\
+✅ Set in the `plugin:regexp/recommended` configuration.\
+🔧 Automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/user-guide/command-line-interface#--fix).\
+💡 Manually fixable by [editor suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).\
+❌ Deprecated.
```
RunDevelopment, line 107: This is not a deal-breaker, but would it be possible to remove the "Deprecated" column? We have all deprecated nodes in their own category, so the column is kinda unnecessary.

Or maybe we should keep it to be more consistent with other projects using ESLint Doc Generator?

bmish: I was thinking the same thing actually so I went ahead and removed the "Deprecated" column using the `ruleListColumns` option.

I also have an issue to consider making this automatic: https://github.com/bmish/eslint-doc-generator/issues/323

RunDevelopment: Ohh, thank you!

RunDevelopment (approved):
LGTM

While I approve, @ota-meshi should have the final say in this. He's the most knowledgeable about the website.

ota-meshi (approved):

PR #510, opened 2023-03-16
Use `@eslint-community` packages

ota-meshi:
This PR will replace the dependency to use the `@eslint-community` package.

FYI, ESLint itself was also recently replaced to use these packages.

https://github.com/eslint/eslint/pull/16784


RunDevelopment (approved):

RunDevelopment:
Ohhh, I didn't know that `eslint-utils` is part of `@eslint-community` as well. I should probably move our more aggressive version of `getStaticValue` to `eslint-utils`.

PR #518, opened 2023-03-30
Updated refa, regex-ast-analysis, and scslre

RunDevelopment:
As the title says, this PR updates my libraries. The main change in all 3 libraries is that I switched to `@eslint-community/regexpp`.

Aside from that, I made several changes to refa, but only a few affect us. The main one would be that DFA/NFA new use node factories to limit the number of nodes created. 

I also changed `getClosestAncestor`, it now takes any number of elements.


ota-meshi (approved):
LGTM! Thank you!

PR #547, opened 2023-08-04
Fix links to `regexp/sort-character-class-elements`

kevinoid:
Fix the links to `regexp/sort-character-class-elements` in the docs for `regexp/order-in-character-class` was were inadvertently linking to `regexp/no-useless-lazy`.


ota-meshi (approved):
Thank you!

PR #548, opened 2023-08-05
Use `changesets`

ota-meshi:
This PR introduces changeset to automate the release process.


RunDevelopment:
Let me start by saying that you do the releasing for this project, so I think that you should feel free to improve the release workflow however you see fit.

That being said, I don't really see what we get from changesets. As someone that has never heard of changesets and has no idea what is does, their documentation didn't exactly help me. Their docs are fairly abstract, have no examples, and constantly refer to elements of their CLI interface, which I can't see because I don't have the CLI open. And their "Explore changelog" link (which I think is an example) is dead. Consequently, my first impression of changesets wasn't the best...

So I have a few questions:

- How does changesets help us?
- Will this change how people have to make PRs?
- Do we have to do anything special when merging PRs?

ota-meshi:
Thank you for your comment!

> How does changesets help us?

I couldn't find any helpful documentation for changeset, but changeset is used by Stylelint and Svelte. I think you can see how these operate.

PRs for feature additions and fixes that will be included in the next release should add files for the changeset. You can add this from changeset-bot's comment.

image

The editor will start with the PR title entered, so enter `patch`, `minor` or `major` and save it.

image

When you merge the PR with these files added, GitHub Action will open the PR prepare for release.
For example the following link is a PR opened in Stylelint.
https://github.com/stylelint/stylelint/pull/7073
We can see what's included in the next release in this PR.

Merging this PR will release the package via GitHub Action.

---

I would like to do a pre-release of eslint-plugin-regexp with support for the v flag and drop support for older Node versions.
I'm hoping that using changesets will make it easier.
https://github.com/changesets/changesets/blob/main/docs/prereleases.md

> Will this change how people have to make PRs?
> Do we have to do anything special when merging PRs?

After creating the PR, we need to add the files for the changeset from the bot's comments. (Not required if the PR is repository maintenance.)
However, it seems that this can also be created from the CLI before opening the PR, but I have not used it much.

RunDevelopment:
I took a look at stylelint, and I think I get it now. The basic idea of `changesets` is that you incrementally build the changelog by adding a changelog entry (=changeset) to every PR. Since each changelog entry is stored as a file in git history, `changesets` can completely generate the changelog by asking git which commit/PR added these files.

It's a clever idea, and I can see it being really useful for mono repos. Having to manually release X-many packages can be a real hassle there.

However, I see an issue with crediting people properly. Maybe I misunderstood, but changes like improving infrastructure or documentation (i.e. everything in our `docs/` folder) wouldn't be mentioned in the changelog, because it's not a patch/minor change, right? If we already used `changesets`, would e.g. #547 appear in the changelog of the next release?

> After creating the PR, we need to add the files for the changeset from the bot's comments.

That bot comment really bugs me. If I was a first-time contributor to stylelint and saw [this comment](https://github.com/stylelint/stylelint/pull/7110#issuecomment-1653079851), I would think that I did something wrong and spend the next hour or two trying to figure out what I did wrong. We already talked about the documentation for `changesets`, so those one or two hours won't be a great time.

It's not clear at all that contributors can completely ignore this comment. Nothing in this comment says we add the changeset. In the end, we (or rather you) decide how to release changes, so why is the first thing contributors see under every PR something completely irrelevant to them? Right now, that message is just confusing for contributors.

Is there a way to change the message of the bot comment? I want to make it clear that contributors don't need to worry about it, and that it is a message for us.

> I would like to do a pre-release of eslint-plugin-regexp with support for the v flag and drop support for older Node versions.
> I'm hoping that using changesets will make it easier.

Sounds good. I never did pre-releases before, but it would be nice if `changesets` could largely/completely automate that process.

ota-meshi:
> Is there a way to change the message of the bot comment? I want to make it clear that contributors don't need to worry about it, and that it is a message for us.

Hmm... I've searched about this but haven't found a way to change the message. Also, after reading the source code of the bot, I don't think we can change the message.

https://github.com/changesets/bot/blob/492290be72c42ae8f9ba85afe636a7b4520a9909/index.ts#L60

RunDevelopment:
Sorry for the delay @ota-meshi!

I still think that `changesets` might make the contributing process unnecessarily complex, but maybe I'm just worrying over nothing. So let's just add `changesets`. If it really does negatively affect us, we can just remove it.

RunDevelopment (approved):

Issue #545, opened 2023-07-22
Support for v flag & update regexpp


ota-meshi:
**Description**

`regexpp` added support for the `v` flag.
Since the AST has changed along with this, many Type errors have occurred. we need to fix them.
Also, some rules need to be fixed to understand the `v` flag.

---

At least the following TS errors should be fixed.

- [x] `regexp/negation` rule
- [x] `regexp/no-contradiction-with-assertion` rule
- [x] `regexp/no-dupe-characters-character-class` rule
- [x] `regexp/no-dupe-disjunctions` rule
- [x] `regexp/no-misleading-capturing-group` rule
- [x] `regexp/optimal-quantifier-concatenation` rule
- [x] `regexp/prefer-character-class` rule
- [x] `regexp/sort-character-class-elements` rule
- [x] `regexp/use-ignore-case` rule
- [x] `lib/utils/index.ts`
- [x] `lib/utils/partial-parser.ts`
- [x] `lib/utils/regexp-ast/case-variation.ts`
- [x] `lib/utils/regexp-ast/index.ts`
- [x] `lib/utils/regexp-ast/is-equals.ts`
- [x] `lib/utils/regexp-ast/simplify-quantifier.ts`

RunDevelopment:
Fixing the type error won't be enough. refa and RAA can't deal with the v flag at all right now. I'll try to add support to both today.

RunDevelopment:
Quick update: I have implemented support for the v flag in refa and RAA. This took a lot longer than expected. It's not released yet, because I still want to test it some more.

Also, while implementing this, the full scope of what we'll have to do dawned on me. We'll probably have to update all rules that interact with characters classes/sets in any way, not just the ones with type errors.

ota-meshi:
In my opinion the `no-empty-alternative` rule should report empty alternatives. e.g. `/[\q{a|}]/v`
Presumably, if we support it, we can say that the existing rules supported the `v` flag. What do you think?

The following rules currently do not have test cases for the `v` flag:

- confusing-quantifier
- control-character-escape
- hexadecimal-escape
- letter-case
- match-any
- no-control-character
- no-empty-alternative
- no-empty-capturing-group
- no-empty-character-class
- no-empty-group
- no-empty-lookarounds-assertion
- no-escape-backspace
- no-extra-lookaround-assertions
- no-invalid-regexp
- no-invisible-character
- no-lazy-ends
- no-legacy-features
- no-missing-g-flag
[...]

PR #554, opened 2023-08-15
Update `@eslint-community/regexpp` to v4.6

ota-meshi:
This PR updates `@eslint-community/regexpp` to v4.6.
TS errors caused by this change are temporarily avoided with suppression comments. We will have to fix these errors later.

Related to #545

### DIFF
--- a/.changeset/small-doors-lie.md
+++ b/.changeset/small-doors-lie.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": minor
+---
+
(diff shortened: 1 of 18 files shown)


RunDevelopment:
> TS errors caused by this change are temporarily avoided with suppression comments. We will have to fix these errors later.

Oh god. This will now be fun. We should probably have an issue to keep track of progress.

RunDevelopment (approved):

PR #557, opened 2023-08-15
Fix `isEqualNodes`

RunDevelopment:
It doesn't work yet, because I need to update refa and RAA. I'll work on that today.


ota-meshi (approved):
LGTM! Thank you!

PR #558, opened 2023-08-15
Drop support for ESLint < v8.44

ota-meshi:
I noticed that ESLint's standard parser can't parse the v flag until ESLint is updated to v8.44.
This PR drops support for ESLint < v8.44.

https://eslint.org/blog/2023/06/eslint-v8.44.0-released/

We could leave the old ESLint support and check the ESLint version, and branch whether or not to test for the v flag, but I think it would be expensive to do that. So I think it makes more sense to drop the old ESLint support.


RunDevelopment:
Why exactly would we need to check the ESLint version and branch? 

The ESLint parser AST represents regexes as `{ pattern: string; flags: string }`, so why does it matter to us whether the parser supports the `v` flag? The only issue I see is that rules that add the `v` flag to regexes (probably a `use-v-flag` rule) will create code that ESLint can't parse. But that's not our problem. We produce valid JS/TS, and users can turn off the rule if it causes issues.

ota-meshi:
The old parser will throw a parsing error with the v flag as an unknown flag. Parsing the v flag requires espree v9.6.0 and is definitely included in eslint v8.44.0 and later.

https://github.com/eslint/espree/releases/tag/v9.6.0

That said, we can certainly use `new RegExp('pattern', 'v')` if we just want to test our rule.

RunDevelopment:
> That said, we can certainly use `new RegExp('pattern', 'v')` if we just want to test our rule.

Oh, so we can't even properly *test* our v-flag rules right now? That certainly is an issue, and I don't like the idea of having to use `RegExp('pattern', 'v')` everywhere. We still won't be able to use regex literals in test, because NodeJS 18 doesn't support the v flag, but `` String.raw`/pattern/v` `` is better than `` String.raw`new RegExp('pattern', 'v')` ``.

PR #560, opened 2023-08-16
Add support for v flag to `regexp/negation` rule

ota-meshi:
related to #545

### DIFF
--- a/lib/rules/negation.ts
+++ b/lib/rules/negation.ts
@@ -59,11 +59,7 @@ export default createRule("negation", {
                         // Note: This only affects Unicode property escapes.
                         // All other character sets are either case-invariant
                         // (/./, /\s/, /\d/) or inconsistent (/\w/).
-
-                        // since we know that the property doesn't contain strings,
-                        // we can just get the characters of the character class
-                        // without worrying about the strings
-                        const ccSet = toUnicodeSet(ccNode, flags).chars
+                        const ccSet = toUnicodeSet(ccNode, flags)
 
                         const negatedElementSet = toCharSet(
                             {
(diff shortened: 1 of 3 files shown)


RunDevelopment:
Relevant: https://bugs.chromium.org/p/chromium/issues/detail?id=1474268

ota-meshi:
Thank you for sharing the link, and making some fixes!
It looks like the Chrome bug you shared hasn't been fixed yet, but I think that ready to review this rule.

RunDevelopment:
The fix for the bug has been merged yesterday and will be part of Chrome 118. 118 will only be released starting on Oct 10 though, so it will take some time until we see the fix. But that's fine, we can release the rule.

lib/rules/negation.ts
```diff
@@ -53,13 +63,12 @@ export default createRule("negation", {
                         // All other character sets are either case-invariant
                         // (/./, /\s/, /\d/) or inconsistent (/\w/).
 
-                        // FIXME: TS Error
-                        // @ts-expect-error -- FIXME
-                        const ccSet = toCharSet(ccNode, flags)
+                        // since we know that the property doesn't contain strings,
+                        // we can just get the characters of the character class
+                        // without worrying about the strings
+                        const ccSet = toUnicodeSet(ccNode, flags).chars
```
RunDevelopment, line 69: I would like to get rid of the `.chars` here. The reasoning explained in the comment isn't trivial, so I would like to simplify it to the code I suggested.

Unfortunately, we can't compare a `UnicodeSet` to a `CharSet` right now, but I'm currently working on that. I'm currently working on a patch for refa and RAA which will include some fixes + more ergonomics for `UnicodeSet`s. So my suggested code will work then.

```suggestion
                        const ccSet = toUnicodeSet(ccNode, flags)
```

RunDevelopment (approved):

PR #568, opened 2023-09-05
Update refa & co

RunDevelopment:
This updates `refa`, `regex-ast-analysis`, and `scslre`.

There are a lot of changes in both refa and RAA because of the `v` flag, so fixing all errors will be fun... I also had to change the semantics of a few common functions in RAA. Noteably, `is{Potentially,}{Empty,ZeroLength}` and a few others now require a flags argument. (Since we can't assume that a single character class/set is a single character anymore, we have to evaluate the whole thing in some cases, and that requires flags.)

This PR is still a work in progress. Since some type errors are caused by the algorithm fundamentally assuming a single character, I will liberally use `// @ts-expect-error -- FIXME` for now.

### DIFF
--- a/.changeset/cyan-waves-notice.md
+++ b/.changeset/cyan-waves-notice.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": minor
+---
+
+Update refa, regexp-ast-analysis, and scslre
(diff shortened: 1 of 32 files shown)


ota-meshi (approved):
LGTM! Thank you for great work!

PR #569, opened 2023-09-06
Use new refa AST transformers

RunDevelopment:
Since the combination of `Transformers.applyAssertions` + `Transformers.removeUnnecessaryAssertions` + everything else was quite common, I added a new transformer for this in refa v0.12.0. This PR just makes 2 rules use the new transformer.

I also fixed `parser.ast.flags.unicode ? 0x10ffff : 0xffff` while I was at it. This was not only wrong because of the `v` flag, but also unnecessary. `parser.maxCharacter` is the value we want.

### DIFF
--- a/.changeset/hungry-spoons-mix.md
+++ b/.changeset/hungry-spoons-mix.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": patch
+---
+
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
LGTM! Thanks!

PR #571, opened 2023-09-06
Fix `parseFlags`

RunDevelopment:
`parseFlags` did not account for the `d` and `v` flags.

This will make the tests pass in #560.


ota-meshi (approved):
Thank you!

PR #575, opened 2023-09-06
Disable `complexity` rule

RunDevelopment:
I don't like this rule. My main issue is that it heavily punishes exhaustive `switch` statements. I use the following pattern all over my TS projects:

```ts
switch (someVar.type) {
  case "a":
  case "b":
    return x;
  case "c":
  case "d":
    return y;
  default:
    return assertNever(someVar);
}
```

([Real example](https://github.com/RunDevelopment/regexp-ast-analysis/blob/008797bb5b9fd9ac92f56f6de0c23d56a706f308/src/basic.ts#L144).) If `type` is the discriminator of some union type, then this pattern ensures that we exhaustively cover all possible variants. So if we later add/remove variants, TS will give us a compiler error. This is great for correctness because it forces us to explicitly state what we want to do for every variant. 

```ts
if (someVar.type == "a" || someVar.type == "b") {
  return x;
} else { 
  return y;
}
```

Compared to the classic `if/else`, the exhaustive `switch` pattern allows TS to catch future errors for us.

But the `complexity` rule punishes this pattern. Every `case` gets for 1 point, so the above `switch` code has a complexity of 4 while the `if/else` has a complexity of 2. Each `case` counting as 1 point is going to be a huge problem for us, because regexpp has 18 node types now, but the maximum allowed complexity is 15. 

---

Side note: I don't like how long the exhaustive `switch` pattern is either, but it's the best we have in JS. What I really want is something like [Rust's `match`](https://doc.rust-lang.org/rust-by-example/flow_control/match.html). Then we could write this:
```
return match (someVar.type) {
  "a" | "b" => x,
  "c" | "d" => y,
  // compiler checks that no variants are missing
};
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
         "require-jsdoc": "error",
         "no-warning-comments": "warn",
         "no-lonely-if": "off",
+        complexity: "off",
         "@typescript-eslint/no-non-null-assertion": "off",
 
         "no-shadow": "off",
--- a/lib/rules/no-dupe-characters-character-class.ts
+++ b/lib/rules/no-dupe-characters-character-class.ts
@@ -255,7 +255,6 @@ export default createRule("no-dupe-characters-character-class", {
             const { flags } = regexpContext
 
             return {
-                // eslint-disable-next-line complexity -- X
                 onCharacterClassEnter(ccNode: CharacterClass) {
                     const {
                         duplicates,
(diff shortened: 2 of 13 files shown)


ota-meshi:
I agree with your opinion. It may be a bad idea to use complexity rule in projects using TypeScript. That can make type guarding more difficult.

ota-meshi (approved):

PR #583, opened 2023-09-09
Improve `regexp/no-invalid-regexp` rule to check for unknown pattern flags.

ota-meshi:
This PR improves the `regexp/no-invalid-regexp` rule to check flags even for unknown patterns.


RunDevelopment (approved):
Nice idea!

PR #584, opened 2023-09-09
Add support for v flag to `regexp/no-misleading-unicode-character` rule

ota-meshi:

### DIFF
--- a/tests/lib/rules/no-misleading-unicode-character.ts
+++ b/tests/lib/rules/no-misleading-unicode-character.ts
@@ -3,7 +3,7 @@ import rule from "../../../lib/rules/no-misleading-unicode-character"
 
 const tester = new RuleTester({
     parserOptions: {
-        ecmaVersion: 2020,
+        ecmaVersion: "latest",
         sourceType: "module",
     },
 })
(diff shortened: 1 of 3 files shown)


tests/lib/rules/no-misleading-unicode-character.ts
```diff
@@ -262,5 +269,52 @@ tester.run("no-misleading-unicode-character", rule as any, {
...
+        },
+        {
+            code: String.raw`/[👶🏻👨‍👩‍👦]/v`,
+            output: String.raw`/[\q{👶🏻|👨‍👩‍👦}]/v`,
+            options: [{ fixable: true }],
+            errors: [{ messageId: "characterClass" }],
+        },
+        {
+            code: String.raw`/[👶🏻&👨‍👩‍👦]/v`,
+            output: String.raw`/[\q{👶🏻|👨‍👩‍👦}&]/v`,
```
RunDevelopment, line 297: Could this potentially create invalid regexes? `\q` isn't allowed in negated character classes, so e.g. `[^👶🏻&👨‍👩‍👦]` can't be fixed to `[^\q{👶🏻|👨‍👩‍👦}&]`. Does this work correctly? If yes, could we add a test for this?

ota-meshi: Thank you for checking it!
I added a test case. It has already been avoided in L137 and seems to be fine.

RunDevelopment (approved):

PR #587, opened 2023-09-09
Improve `regexp/sort-alternatives` rule to add support for string alternatives and v flag

### LINKED ISSUE #572: Sort string alternatives
The order of string alternatives (e.g. `\q{foo|bar}`) does not matter, so it would be nice if they could automatically be sorted. The semantics and functionality should be the same as [`sort-alternatives`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/sort-alternatives.html). 

I'm not sure whether sorting string alternatives should be a new rule or whether it should be part of `sort-alterantives`.

ota-meshi:
close #572

### DIFF
--- a/lib/rules/sort-alternatives.ts
+++ b/lib/rules/sort-alternatives.ts
@@ -6,6 +6,7 @@ import type {
     CharacterSet,
     ClassStringDisjunction,
     Element,
+    ExpressionCharacterClass,
     Pattern,
     StringAlternative,
 } from "@eslint-community/regexpp/ast"
@@ -33,6 +34,7 @@ import {
     getLongestPrefix,
     getConsumedChars,
     toUnicodeSet,
+    hasStrings,
 } from "regexp-ast-analysis"
 import type { CharSet, Word, ReadonlyWord } from "refa"
 import { NFA, JS, transform } from "refa"
@@ -88,7 +90,8 @@ function containsOnlyLiterals(
                 d.type === "Backreference" ||
                 d.type === "CharacterSet" ||
                 (d.type === "Quantifier" && d.max === Infinity) ||
-                (d.type === "CharacterClass" && d.negate)
+                (d.type === "CharacterClass" && d.negate) ||
+                (d.type === "ExpressionCharacterClass" && d.negate)
             )
         },
         (d) => d.type !== "Assertion",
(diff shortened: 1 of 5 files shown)


lib/rules/sort-alternatives.ts
```diff
@@ -213,11 +233,16 @@ function getLexicographicallySmallestFromAlternative(
...
-): nodes is readonly (Character | CharacterClass | CharacterSet)[] {
-    return nodes.every(
+/**
+ * Returns whether the given alternative has contains only characters.
+ * But note that if the pattern has the v flag, the character class may contain strings.
+ */
+function hasOnlyCharacters(
+    alternative: Alternative,
+): alternative is Alternative & {
+    elements: readonly (Character | CharacterClass | CharacterSet)[]
```
RunDevelopment, line 243: Expression character class is missing.

lib/rules/sort-alternatives.ts
```diff
@@ -157,25 +158,44 @@ function approximateLexicographicallySmallest(
...
+            const firstCharSets = [
+                ...(cs.chars.isEmpty ? [] : [cs.chars]),
+                ...cs.accept.wordSets
+                    .map((ws) => (ws.length ? ws[0] : null))
+                    .filter((wcs): wcs is CharSet => Boolean(wcs)),
+            ]
+            if (!firstCharSets.length) return undefined
+            smallest.push(
+                Math.min(...firstCharSets.map((fcs) => fcs.ranges[0].min)),
+            )
```
RunDevelopment, line 198: This isn't correct. Due to `ws[0]`, this function will incorrectly say that the lexicographically smallest word of `a[\q{bc}]d` is `abd`.

Doing this correctly isn't easy, because we have to account for the next characters in the alternative. We can't just pick the lexicographically smallest word in the current Unicode set. E.g. the lexicographically smallest word in `[\q{a|aba}]` is `a`, but the lexicographically smallest word `[\q{a|aba}]bb` is `ababb`.  So we have to construct the smallest word by going through the elements back to front.

Of course, if the current Unicode set doesn't contain strings, then we don't have to account for the characters after it, so we can optimize the common case.

ota-meshi: Thank you for checking this PR!
I tried to fix the problem. Could you please check again?

lib/rules/sort-alternatives.ts
```diff
@@ -156,29 +163,72 @@ function approximateLexicographicallySmallest(
...
+        const smallest: Word = []
+        for (const e of reversedElements) {
+            const us = toUnicodeSet(e, flags)
+            if (us.isEmpty) return undefined
+            if (us.accept.isEmpty) {
+                smallest.unshift(us.chars.ranges[0].min)
+            } else {
+                const words: ReadonlyWord[] = [
+                    ...(us.chars.isEmpty ? [] : [[us.chars.ranges[0].min]]),
+                    ...us.accept.words,
```
RunDevelopment, line 210: Same issue as in #588. Fixed with #604.

ota-meshi: I've added the same fix to this PR. (But there may be conflicts)

ota-meshi:
@RunDevelopment  I fixed getLexicographicallySmallestInConcatenation a little. Could you check if it is the correct fix?
https://github.com/ota-meshi/eslint-plugin-regexp/pull/587/commits/b0ddd9f4c2747597eaa640b04d093163fbdec102

RunDevelopment (approved):
Thank you @ota-meshi and sorry for the delay!

PR #588, opened 2023-09-09
Add support for v flag to `regexp/sort-character-class-elements` rule

ota-meshi:


tests/lib/rules/sort-character-class-elements.ts
```diff
@@ -230,5 +233,57 @@ tester.run("sort-character-class-elements", rule as any, {
...
+        {
+            code: String.raw`/[[a]a\q{a}[a--b]]/v`,
+            output: String.raw`/[\q{a}[a]a[a--b]]/v`,
+            errors: [
+                "Expected character class elements to be in ascending order. '\\q{a}' should be before '[a]'.",
+            ],
+        },
+        {
+            code: String.raw`/[\q{a}[a]a[a--b]]/v`,
+            output: String.raw`/[\q{a}[a][a--b]a]/v`,
```
RunDevelopment, line 253: Is this really the order we should go for? I would have expected the order to be `[a\q{a}[a][a--b]]`.

I think that the order `*` > `\q` > `[]` might be better, because it prevents cases like `[\q{a}[a][a--b]a]`. That tiny last `a` is easy to miss, but that's not a problem with `[a\q{a}[a][a--b]]` IMO.

Thoughts?

ota-meshi: Thank you for your thoughts!
I think the order you suggested is easier to read. I changed the default order.

lib/rules/sort-character-class-elements.ts
```diff
@@ -37,9 +46,53 @@ function getCharacterClassElementKind(
...
+    node: CharacterClassElement,
+    flags: ReadonlyFlags,
+): Word {
+    const us =
+        node.type === "CharacterSet" && node.negate
+            ? toUnicodeSet({ ...node, negate: false }, flags)
+            : toUnicodeSet(node, flags)
+    const wordSets: ReadonlyWordSet[] = [
+        ...(us.chars.isEmpty ? [] : [[us.chars]]),
+        ...(us.accept.isEmpty ? [] : us.accept.wordSets),
```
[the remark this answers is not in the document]
ota-meshi: I further checked the code and it seemed more efficient to use words than wordSets, so I refactored it further.

https://github.com/ota-meshi/eslint-plugin-regexp/pull/588/commits/e997b3f3fb0e4c37ef2e375d64f16ae35a5c3bde

lib/rules/sort-character-class-elements.ts
```diff
@@ -37,9 +46,50 @@ function getCharacterClassElementKind(
...
+    node: CharacterClassElement,
+    flags: ReadonlyFlags,
+): ReadonlyWord {
+    const us =
+        node.type === "CharacterSet" && node.negate
+            ? toUnicodeSet({ ...node, negate: false }, flags)
+            : toUnicodeSet(node, flags)
+    const minimumWords: ReadonlyWord[] = [
+        ...(us.chars.isEmpty ? [] : [[us.chars.ranges[0].min]]),
+        ...us.accept.words,
```
RunDevelopment, line 75: Uhhhmmm, my bad. `StringSet#words` wasn't meant to be public, and I made it private in the last update. The code still technically works, because the field is still there, but we really shouldn't rely on it, because I will probably remove it in the future.

ota-meshi: Thanks for fixing it!

RunDevelopment (approved):

PR #578, opened 2023-09-06
`prefer-w`: Add support for `v` flag

RunDevelopment:
Besides adding support for the `v` flag, I also fixed a bug. `\D` was incorrectly detected as `\d` in `isDigitRangeOrSet`.

PR #590, opened 2023-09-10
Add support for v flag to `regexp/prefer-w` rule

ota-meshi:


ota-meshi:
I just realized that work has already started 😅  #578

PR #595, opened 2023-09-11
Add `regexp/simplify-set-operations` rule

### LINKED ISSUE #574: Simplify nested characters classes and expression character classes
**Motivation**
The `v` flag added set operation to character classes. These set operations can be simplified in some cases.

**Description**
The rule should apply the following simplifications:

1. `[[abc]]` -> `[abc]` and `[^[abc]]` -> `[^abc]` and `[[^\w&&\d]]` -> `[^\w&&\d]`: remove trivially nested (expression) character classes.
1. `[[abc][def]ghi]` -> `[abcdefghi]`: always inline non-negated character classes.
1. `[a&&[^b]]` -> `[a--b]` and `[[^b]&&a]` -> `[a--b]`: use subtraction
1. `[a--[^b]]` -> `[a&&b]`: use intersection
1. `[[^a]&&[^b]]` -> `[^ab]`: De Morgan

ota-meshi:
close #574

I found a bug in `regexpp` while implementing this rule. We need to fix that first.

https://github.com/eslint-community/regexpp/pull/136

If you have an idea for a better rule name, please let me know. I can rename the rule name.


ota-meshi:
Now that regexpp has been fixed, I think it's ready for review.

tests/lib/rules/require-reduce-negation.ts
```diff
@@ -0,0 +1,200 @@
...
+        },
+        {
+            code: String.raw`/[^\D]/u`,
+            output: String.raw`/[\d]/u`,
+            errors: [
+                "This character class can be double negation elimination.",
+            ],
+        },
+        {
+            code: String.raw`/[^\P{ASCII}]/u`,
```
RunDevelopment, line 42: Could you please add a test of this, but with the `i` flag enabled? I want to test it know that `/[\p{ASCII}]/iu` != `/[^\P{ASCII}]/iu`

docs/rules/require-reduce-negation.md
```diff
@@ -0,0 +1,92 @@
...
+This rule is aimed at optimizing patterns by reducing the negation (complement) representation of character classes (with `v` flag).
+
+<eslint-code-block fix>
+
+```js
+/* eslint regexp/require-reduce-negation: "error" */
+
+/* ✗ BAD */
+var re = /[^[^abc]]/v; // -> /[abc]/v
+var re = /[^\D]/u; // -> /[\d]/u
```
RunDevelopment, line 28: So this rule is a superset of [`regexp/negation`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/negation.html)? I'm not quite sure what to do here. I think that it makes sense for `regexp/require-reduce-negation` to fix this issue, but then why does ``regexp/negation`` even exist?

I think we have a few choices here:

1. Just let it be. We'll leave both rules as is and think of `regexp/negation` as a simpler `regexp/require-reduce-negation`.
2. We remove `regexp/negation`.
3. We restrict `regexp/require-reduce-negation` to only fix issues that `regexp/negation` does not fix.

RunDevelopment:
Also, the code for getting escapes right when inlining character classes looks very complex. Maybe we could say that this rule only removes negations and doesn't inline character classes? E.g. `[[^a][^b]]` -> `[^[a][b]]`. We would leave the inlining to `no-useless-character-class`. What do you think?

docs/rules/require-reduce-negation.md
ota-meshi: Hmm..  I think the original purpose of the `negation` rule is to rewrite a pattern that everyone considers meaningless into a correct pattern. For example, in JS, we would like to prohibit things like `if(!true)` and `if(!false)`.
I think the new rule should be main purpose at pattern optimization.
Therefore, in my opinion, I would like to make the two rules separate.

I think that the pattern `/[^[^abc]]/v;` might also be reported in the `negation` rule, but I'm not sure if new rule need to ignore it. Because it is reported both for optimization and for finding mistakes.

ota-meshi:
> We would leave the inlining to `no-useless-character-class`. What do you think?

I think that's a good idea! I will change the rule.

docs/rules/require-reduce-negation.md
RunDevelopment: I like this idea.

As for `negation`, I think we can justify the new functionality by saying that `negation` applies the negation to the character class element. So for some `[^x]`, it will always try to negate `x` (no matter what that `x` is). This would justify `[^[a]]` -> `[^a]`, `[^[^a]]` -> `[a]`, `[^\s]` -> `[\S]` -> `\S`, and `[^[a&&b]]` -> `[^a&&b]`.

As for `De Morgan` and `Subtraction to intersection, and intersection to subtraction`, they might have to be one rule. The issue is that `Subtraction to intersection, and intersection to subtraction` might want to convert `[^a]&&[^b]` to `[^a]--b`. So maybe the combination of the two could be this rule?

ota-meshi: > The issue is that Subtraction to intersection, and intersection to subtraction might want to convert [^a]&&[^b] to [^a]--b. So maybe the combination of the two could be this rule?

Ah, that's true. If we add two rules, the rules may conflict. It makes sense to add one rule. I have no idea for the name of the new rule. Do you have any ideas for good rule name?

RunDevelopment: Maybe something like `simplify-set-operations`?

ota-meshi: Thanks for sharing your idea! I'll rename the rule to that name later.

ota-meshi: I changed the rule name and changed the docs.

docs/rules/simplify-set-operations.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### Conversion from the subtraction to the intersection
+
+Difference set with a complement operand on the right side can be converted to intersection sets.\
+The rule looks for character class subtraction with negation operand on the right side, reports on them, auto-fix them.\
+For example, `/[a--[^b]]/v` is equivalent to `/[a&&b]/v`.
+
+### Auto Fixes
+
+This rule's auto-fix does not remove unnecessary brackets. For example, `/[[^a]&&[^b]]/v` will be automatically fixed to `/[[a][b]]/v`.\
```
RunDevelopment, line 70: Let's hope that it won't be auto-fixed to that :)

```suggestion
This rule's auto-fix does not remove unnecessary brackets. For example, `/[[^a]&&[^b]]/v` will be automatically fixed to `/[^[a][b]]/v`.\
```

docs/rules/simplify-set-operations.md
```diff
@@ -0,0 +1,92 @@
...
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+> require simplify set operations
+
+## :book: Rule Details
+
+This rule aims to optimize patterns by simplifying set operations on character classes (with `v` flag).
```
RunDevelopment, line 19:
```suggestion
This rule aims to optimize patterns by simplifying set operations in character classes (with `v` flag).
```

docs/rules/simplify-set-operations.md
```diff
@@ -0,0 +1,92 @@
...
+
+This rule does not report simple nested negations. (e.g. `/[^[^abc]]/v`)\
+If you want to report simple nested negations, use [regexp/negation] rule together.
```
RunDevelopment, line 22:
```suggestion
If you want to report simple nested negations, use the [regexp/negation] rule.
```

docs/rules/simplify-set-operations.md
```diff
@@ -0,0 +1,92 @@
...
+
+</eslint-code-block>
+
+### How does this rule work?
+
+This rule attempts to simplify set operations in the ways listed below:
+
+#### De Morgan's laws
+
+This rule uses De Morgan's laws to look for patterns that can convert multiple negations into a single negation, reports on them, auto-fix them.\
```
RunDevelopment, line 51:
```suggestion
This rule uses De Morgan's laws to look for patterns that can convert multiple negations into a single negation, reports on them, and auto-fix them.\
```

docs/rules/simplify-set-operations.md
```diff
@@ -0,0 +1,92 @@
...
+For example, `/[[^a]&&[^b]]/v` is equivalent to `/[^ab]/v`, `/[[^a][^b]]/v` is equivalent to `/[^a&&b]/v`.
```
RunDevelopment, line 52:
```suggestion
For example, `/[[^a]&&[^b]]/v` is equivalent to `/[^ab]/v`, and `/[[^a][^b]]/v` is equivalent to `/[^a&&b]/v`.
```

lib/rules/simplify-set-operations.ts
```diff
@@ -0,0 +1,419 @@
...
+            /**
+             * Checks the given character class and reports if it can be converted to the negation of a conjunction
+             * using De Morgan's laws.
+             * Returns true if reported.
+             *
+             * e.g.
+             * - `[[^a][^b]]` -> `[^a&&b]`
+             */
+            function toNegationOfConjunction(ccNode: CharacterClass) {
+                if (ccNode.elements.length <= 1 || !flags.unicodeSets) {
```
RunDevelopment, line 267: I think this flag check is redundant.

```suggestion
                if (ccNode.elements.length <= 1) {
```

lib/rules/simplify-set-operations.ts
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        return defineRegexpVisitor(context, {
+            createVisitor,
+        })
+    },
+})
+
+/** Gets the parsed result element. */
+function getParsedElement(
```
RunDevelopment, line 389: Could we move this function up before `export default createRule(...)`? We have `export default createRule(...)` as the last statement in all other rules (I think), so I would like to keep it consistent.

ota-meshi: Yeah. I need to move it up.

RunDevelopment (approved):
Thank you @ota-meshi!

PR #598, opened 2023-09-12
Add `regexp/require-unicode-sets-regexp` rule

### LINKED ISSUE #582: Add `require-unicode-sets-regexp`
Same as the [`require-unicode-regexp`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/require-unicode-regexp.html) rule, but for the `v` flag.

Auto fixing should work the same as for `require-unicode-regexp`. We will only have a fix if we can prove that adding the `v` flag doesn't change the behavior of the regex.

Note: It might be difficult to migrate from legacy (no `u` or `v` flag) to the `v` flag in one step. We should probably migrate the `u` (if not present already) first.

ota-meshi:
close #582


RunDevelopment:
The compatibility check isn't enough, because the `v` might cause the regex to be parsed differently. E.g. `/[a&&b]/u`. Of course, edge cases like this are going to be extremely rare, since e.g. `[a&&b]` doesn't make sense without the `v` flag. 

Fixing this should be relatively easy. We just to verify that no character classes in the `u` regex contain any double punctuation sequences in their raw source code.

ota-meshi:
Thank you for checking this PR. I fixed the problem. Could you please check again?

RunDevelopment (approved):

PR #599, opened 2023-09-13
Update refa and RAA

RunDevelopment:
Changes:
- I added a lot of functionality to `UnicodeSet`. We should now have an easier time adding support for the `v` flag, since `UnicodeSet` is a lot more similar to `CharSet` now. Also, bug fixes.
- Some minor bug fixes in RAA.


ota-meshi (approved):
Thank you!

PR #601, opened 2023-09-13
Disable `require-jsdoc` rule

### LINKED ISSUE #577: Disable `require-jsdoc` rule?
I would like to disable the `require-jsdoc` ESLint rule. I don't like this rule, because it encourages code like this:

```ts
/** Extract property references from the given pattern */
export function* extractPropertyReferencesForPattern(
    node: Pattern,
    context: Rule.RuleContext,
): Iterable<PropertyReference> {
    yield* iteratePropertyReferencesForPattern(node, context)
}
```

What is the point of having a JSDoc comment when the comment is just the function name + arguments restated as a sentence? These JSDoc comments do not add information. Reading them is a waste of time, because we can get more information by looking at the function signature.

We all agree that littering our code with comments like this is bad:

```ts
// decrement i
i--
```

So why do we have an ESLint rule that encourages the JSDoc equivalent of this? 

---

So yeah. That's why I would like to disable this rule. I would also like to remove the ~200 useless doc comments scattered throughout the entire code base. This would likely change almost every file (e.g. just look at the few dozen `/** Create visitor */ function createVisitor`). 

What do you think @ota-meshi?

RunDevelopment:
Resolves #577.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,7 @@ module.exports = {
         "plugin:regexp/recommended",
     ],
     rules: {
-        "require-jsdoc": "error",
+        "require-jsdoc": "off",
         "no-warning-comments": "warn",
         "no-lonely-if": "off",
         complexity: "off",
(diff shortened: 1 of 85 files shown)


ota-meshi (approved):
Thank you!

PR #602, opened 2023-09-14
Add support for v flag to `regexp/prefer-d` rule

ota-meshi:


RunDevelopment (approved):
Nice one!

PR #607, opened 2023-09-14
Use `assertNever`

RunDevelopment:
I moved `assertNever` into its own util file. 

The file is called `util.ts`, because I intend to move other general utility function into it (= function and constants that generally just fill holes in the language). Here's an example of this file in 2 other projects I work on: [Navi](https://github.com/chaiNNer-org/Navi/blob/main/src/util.ts), [refa](https://github.com/RunDevelopment/refa/blob/master/src/util.ts). This file will never have any imports, because it only depends on JavaScript itself.

I also used `assertNever` in a few places where it wasn't before.


ota-meshi (approved):
Thank you!

PR #616, opened 2023-09-24
Add `regexp/prefer-set-operation` rule

### LINKED ISSUE #581: Use expression character class instead of lookaround
**Motivation**
Due to the lack of set operation for character classes, I wrote regexes like `(?!\s)[\w\P{ASCII}]` in the past ([another example](https://github.com/ota-meshi/eslint-plugin-regexp/issues/574#issuecomment-1711508301)). They can be transformed to use set operations instead. Using set operation not only makes the intent clearer, it's also more performant. I did some performance test a while back, and `(?!\s)[\w\P{ASCII}]` was about 4x slower than `[\w...whatever ranges are equivalent]`.

**Description**
Combine single-character lookarounds with character classes in regexes with the `v` flag. E.g. `(?=[x])[y]` => `[y&&x]` and `(?![x])[y]` => `[y--x]`. Regexes without the `v` should not be affected.

If the element after the lookaround is a character set, then the transformation should still be applied. E.g. `(?!\s)\P{ASCII}` => `[\P{ASCII}--\s]`.

**Examples**

```js
// good
/(?!\s)[\w\x80-\uFFFF]/;
/(?!\s)[\w\P{ASCII}]/u;

// bad
bad  = /(?!\s)[\w\P{ASCII}]/v;
good = /[[\w\P{ASCII}]--\s]/v;
```

RunDevelopment:
Resolves #581.

The implementation of the rule is fairly simple right now. It just goes through all elements of an alternative and tries to find the pattern `(?=a)b` or `a(?<=b)`. So it won't attempt to do clever things like assertions reordering (e.g. `(?!a)(?=bcd)e` == `(?=bcd)(?!a)e` == `(?=bcd)[e--a]`) and partial application (e.g. `(?!a|bcd)e` == `(?!bcd)(?!a)e` == `(?!bcd)[e--a]`). I think it should be okay. Adjacent assertions are rare in practice, and partial application has the issue that it might make the intent of the author less clear.


ota-meshi (approved):
LGTM! Thank you

PR #617, opened 2023-09-26
Add support for `v` flag to `regexp/use-ignore-case`

RunDevelopment:
Related to #545.

Changes:
- The rule (and `isCaseVariant`) now supports the `v` flag.
- I added support for including string alternatives in the analysis.
- Fixed `removeAll`. That function did now work correctly because `pattern = pattern.slice(0, start) + pattern.slice(end)` is only correct for a `{ start, end }` pair.
- The rule will now add the `i` in the correct place. E.g. `u` -> `iu`.
- I added a `cachedFn` util function and used it in 3 places. The function itself is very simple.

### DIFF
--- a/.changeset/cyan-rats-attend.md
+++ b/.changeset/cyan-rats-attend.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": minor
+---
+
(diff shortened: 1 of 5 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #619, opened 2023-09-26
Add support for `v` flag to `regexp/prefer-character-class`

RunDevelopment:
Related to #545

Changes:
- What the title says.
- Added `unicode-set.ts` util file for reserved punctuators.
- Changed `no-useless-character-class`, `no-useless-escape`, and `require-unicode-sets-regexp` to use the new util file.

Some details about how I changed `prefer-character-class`:
- This rule assumes that character elements are exactly one character long. So it's hard to add support for e.g. `\q{abc}`. So I restricted it to character elements that don't contain strings. So the rule won't combine e.g. `[abc]|[\q{abc}]` -> `[abc\q{abc}]`. I think support for this can be added in a follow-up PR.
- Since the parsing with the `v` flag is stricter, I had to change `elementsToCharacterClass` (the function that creates the combined character class) to make more escapes. Those additional escapes apply to all regexes regardless of flags.

### DIFF
--- a/.changeset/clean-kids-mate.md
+++ b/.changeset/clean-kids-mate.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": minor
+---
+
+Add support for `v` flag to `regexp/prefer-character-class`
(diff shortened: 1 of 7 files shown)


ota-meshi (approved):
LGTM! Thank you! I'm glad that the scattered code has been organized 😄

PR #621, opened 2023-09-26
Add suggestions for `regexp/no-empty-alternative`

RunDevelopment:
I added suggestions for `no-empty-alternative` that replace the empty alternative with a quantifier. This will hopefully make the rule easier to use.


ota-meshi (approved):
LGTM! I think that's a great suggestion!

PR #624, opened 2023-09-26
Add suggestions for `regexp/no-lazy-ends`

RunDevelopment:
I added 2 suggestions to every report:

1. The first will transform the regex into what it is. E.g. `/a+?/` -> `/a/`, because these 2 regexes are equivalent.
2. The second will make the quantifier greedy. E.g. `/a+?/` -> `/a+/`. I thought that this might be useful.


RunDevelopment:
Okay, how should I fix this? Our tests seem to require that all messages end with a dot, but my suggestion messages end with a sentence in parentheses, so it ends with `.)`.

ota-meshi:
That message should be allowed, so I fixed the testcase for messages :smile:

ota-meshi (approved):
LGTM! Thank you!

PR #625, opened 2023-09-28
Add `regexp/no-useless-set-operand` rule

### LINKED ISSUE #573: Remove unnecessary elements in class intersections and subtractions
**Motivation**
The `v` flags adds class intersection and subtraction as new features. They are simple set operations but they also make it possible to have completely useless elements.

**Description**
Report useless elements in class intersections and subtractions.

**Examples**

```js
/* ✗ BAD */
var foo = /[\w&&\d]/v // => /[\d]/v
// \d is a subset of \w, so remove the intersection.
var foo = /[\w&&\s]/v // => /[]/v
// \w and \s are disjoint, so the intersection will be empty.
var foo = /[\w&&[\d\s]]/v // => /[\w&&[\d]]/v
// \s in [\d\s] and \w are disjoint, so \s can be removed.
// Proof: [\w&&[\d\s]] == [[\w&&[\d]][\w&&\s]] == [[\w&&[\d]][]] == [\w&&[\d]]
var foo = /[\w&&[^\d\s]]/v // => /[\w&&[^\d]]/v
// \s in [^\d\s] and \w are disjoint, so \s can be removed.
// Proof: [\w&&[^\d\s]] == [\w&&[[^\d]&&[^\s]]] == [\w&&[[^\s]&&[^\d]]] == [[\w&&[^\s]]&&[^\d]] == [[\w]&&[^\d]]

var foo = /[\w--\s]/v // => /[\w]/v
// \w and \s are disjoint, so the subtraction is useless.
var foo = /[\d--\w]/v // => /[]/v
// \d is a subset of \w, so the subtraction will be empty.
var foo = /[\w--[\d\s]]/v // => /[\w--[\d]]/v
// \s in [\d\s] and \w are disjoint, so \s can be removed.
// Proof: [\w--[\d\s]] == [[\w--\s]--[\d]] == [\w--[\d]]
var foo = /[\w--[^\d\s]]/v // => /[\w--[^\d]]/v
// \s in [^\d\s] and \w are disjoint, so \s can be removed.
// Proof: [\w--[^\d\s]] == [\w--[[^\d]&&[^\s]] == [[\w--[^\d]][\w--[^\s]]] == [[\w--[^\d]][]] == [\w--[^\d]]
```

RunDevelopment:
Resolves #573.

This implements the rules I described in #573.


lib/rules/no-useless-set-operand.ts
```diff
@@ -0,0 +1,242 @@
...
+
+export default createRule("no-useless-set-operand", {
+    meta: {
+        docs: {
+            description:
+                "disallow unnecessary elements in expression character classes",
+            category: "Best Practices",
+            // TODO Switch to recommended in the major version.
+            // recommended: true,
+            recommended: false,
```
ota-meshi, line 76: We are currently working on a major version update so we can include this rule in recommended ruleset.

```suggestion
            recommended: true,
```

.changeset/nervous-yaks-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-regexp": patch
```
ota-meshi, line 2: I don't think this is a patch. Probably major because it changes the recommended configuration.

```suggestion
"eslint-plugin-regexp": major
```

ota-meshi (changes requested):

lib/rules/no-useless-set-operand.ts
```diff
@@ -0,0 +1,242 @@
...
+    return []
+}
+
+function removeDescendant(root: Node, e: FlatElement): string {
+    let { start, end } = e
+
+    if (e.type === "StringAlternative") {
+        if (e.parent.alternatives.length === 1) {
+            // we have to remove the whole string disjunction
+            // eslint-disable-next-line no-param-reassign -- x
```
ota-meshi, line 49: If you wish to turn off this rule for the entire repository, we can do so.
The ruleset is primarily intended for repositories that I maintain alone.

RunDevelopment: No, I actually like this rule. It enforces that parameters are `const`, which I think is a good idea. It's just that sometimes (rarely) I need parameters to be `let`.

ota-meshi (approved):
LGTM! Thank you!

PR #628, opened 2023-10-01
Add support for `v` flag to `regexp/match-any`

RunDevelopment:
Changes:
- Fixed that character classes with strings were lost their strings after fixing. E.g. `[\q{abc}\s\S]` -> `[\s\S]`.
- Check expression character classes as well.


ota-meshi (approved):
LGTM! Thank you!

PR #642, opened 2023-10-03
Use Intl.Segmenter instead of grapheme-splitter

ota-meshi:


lib/utils/extract-capturing-group-references.ts
```diff
@@ -184,6 +184,13 @@ const WELL_KNOWN_ARRAY_METHODS: {
     flat: {},
     // ES2022
     at: { result: "element" },
+    // ES2023
+    findLast: { elementParameters: [0], result: "element" },
+    findLastIndex: { elementParameters: [0] },
+    toReversed: { result: "array" },
+    toSorted: { elementParameters: [0, 1], result: "array" },
+    toSpliced: { result: "array" },
+    with: { result: "array" },
```
ota-meshi, line 193: In order to use Intl.Segmenter, it is necessary to specify es2023 in the tsconfig `lib`, and I fixed the type error that occurred due to that.

RunDevelopment:
[Firefox doesn't support `Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter#browser_compatibility), so this will likely break the playground on our website on Firefox. There is [a polyfill](https://www.npmjs.com/package/intl-segmenter-polyfill), but it uses WASM, which might be difficult to integrate it (especially since WASM loads asynchronously). They [plan to add support](https://bugzilla.mozilla.org/show_bug.cgi?id=1423593) for it soon-ish though ("We're planning this for Q4").

RunDevelopment (approved):

PR #646, opened 2023-10-04
Add `regexp/grapheme-string-literal` rule

### LINKED ISSUE #644: Force one grapheme in the disjunction of string literals
**Motivation**

It may be useful to be able to force the disjunction of string literals to use only strings that humans recognize as single characters.

**Description**

It is possible to use strings of two or more characters in the disjunction of string literals, but normal disjunction is also possible. (However, there are differences in the evaluation order.)
Forcing a single character to be used in the disjunction of string literals allows people to distinguish between uses.

I think it is better to implement the rule by using `Intl.Segmenter` and allowing it if it is recognized as one grapheme.

**Examples**

```js
/* ✓ GOOD */
var foo = /[\p{RGI_Emoji}--\q{🇦🇨|🇦🇩|🇦🇪|🇦🇫|🇦🇬|🇦🇮|🇦🇱|🇦🇲|🇦🇴|🇦🇶|🇦🇷|🇦🇸|🇦🇹|🇦🇺|🇦🇼|🇦🇽|🇦🇿|🇧🇦|🇧🇧|🇧🇩|🇧🇪|🇧🇫|🇧🇬|🇧🇭|🇧🇮|🇧🇯}]/v

/* ✗ BAD */
var foo = /[\q{abc|def}]/v
```

ota-meshi:
close #644


lib/utils/type-tracker/type-data/string.ts
```diff
@@ -132,6 +132,8 @@ const getPrototypes: () => {
         replaceAll: RETURN_STRING,
         // ES2022
         at: RETURN_STRING,
+        // ES2021
+        replaceAll: RETURN_STRING,
```
ota-meshi, line 136:
```suggestion
```

tests/lib/rules/grapheme-string-literal.ts
```diff
@@ -0,0 +1,209 @@
...
+
+        String.raw`/[\q{竈|門|禰󠄀|豆|子}\q{煉󠄁|獄|杏|寿|郎}]/v`,
+        String.raw`/[\q{a|b|c|}]/v`,
+    ],
+    invalid: [
+        {
+            code: String.raw`/[\q{abc}]/v`,
+            errors: [
+                {
+                    message: "Use single grapheme in string literal.",
```
RunDevelopment, line 182: The error message needs to be improved. If I didn't read the docs for the rule, I would have no idea what the error message wants me to do.

Maybe we could say this: "Only single characters and graphemes are allowed inside character classes. Use regular alternatives (e.g. `(?:foo|[...])`) for strings instead."

lib/rules/grapheme-string-literal.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { RegExpContext } from "../utils"
+import { createRule, defineRegexpVisitor } from "../utils"
+
+const segmenter = new Intl.Segmenter()
+
+export default createRule("grapheme-string-literal", {
+    meta: {
+        docs: {
+            description: "enforce single grapheme in string literal",
+            category: "Best Practices",
```
RunDevelopment, line 11: I'm not sure whether this should be a best practice. The practice this rule enforces is good, because it brings regexes into a canonical form (e.g. don't use `[\q{foo|bar}]`, use `(?:foo|bar)`), but that's more of a stylistic thing IMO.

tests/lib/rules/grapheme-string-literal.ts
ota-meshi: Thank you! I will use that message.

lib/rules/grapheme-string-literal.ts
ota-meshi: I agree with you 👍  I will categorize it under "Stylistic Issues".

RunDevelopment (approved):

Issue #649, opened 2023-10-07
Prepare for v2 release.


ota-meshi:
I think we are ready to release v2.
However, just to be sure, I would like to check to see if there is really any remaining work.

- [x] Are there any other rules that should be added to the recommended set?
- [x] Are there any old ones to drop?

ota-meshi:
I think we need to replace jsdoctypeparser with jsdoc-type-pratt-parser. It seems that eslint-plugin-jsdoc was also replaced quite a while ago.
https://www.npmjs.com/package/jsdoc-type-pratt-parser

ota-meshi:
@RunDevelopment 
I noticed that [no-super-linear-move] is not included in the recommended config.
I would like to include this in the recommended config, what do you think?
[no-super-linear-move]: https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-super-linear-move.html

RunDevelopment:
> Are there any other rules that should be added to the recommended set?

Nope.

> Are there any old ones to drop?

We have a few deprecated ones. We might as well remove them for good.

> I noticed that [no-super-linear-move](https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-super-linear-move.html) is not included in the recommended config.
> I would like to include this in the recommended config, what do you think?

Not a good idea, unfortunately. The issue is that most cases of super linear moves simply cannot be fixed. You would need to switch to a different regex engine... This isn't an option most of the time, so I did not include the rule in recommended.

RunDevelopment:
I guess we're in a pretty good spot to release v2.

But I would let it sit for a week or two and only release 2.0.0 then. Maybe there's still something we'll find.

ota-meshi:
That's a good idea. I'll prerelease again and wait a week.

RunDevelopment:
Alright, I think we're good to go. 

I thought of 2 things that we could do to improve the new rules, but it turned out that we already did both of them, so that's good :) Now my only worry is that people might have trouble upgrading to 2.0.0, because we increased our requirements by a lot (nodejs, ESLint, `Intl.Segement` and Firefox). But this is an issue that will solve itself with time, so we probably don't have to worry about it.

So I think we're ready.

ota-meshi:
I will release v2. If your project requires support for older ESLint or Nodejs, I think we can add support for them later. If so, please let me know.

ota-meshi:
I just released v2 🎉 
[...]

PR #651, opened 2023-10-07
Use jsdoc-type-pratt-parser instead of jsdoctypeparser

ota-meshi:
https://github.com/ota-meshi/eslint-plugin-regexp/issues/649#issuecomment-1751716406


RunDevelopment (approved):
Code looks good and tests pass.

PR #654, opened 2023-10-08
Use new ESLint API

ota-meshi:
This PR refactors to use the new API instead of the old ESLint API.

See https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/

They are also present in the minimum version we support.


RunDevelopment:
Code looks good, but I'm a bit worried about updating `eslint` and `@types/eslint`. Wouldn't this enable us to use newer API functions than our minimum supported ESLint version? I'm not sure if this will be an issue, but ideally we could use TS to enforce our minimum supported ESLint version. What do you think?

ota-meshi:
I think we should use `eslint@8.44.0` as you say :+1:
I made sure that `eslint@8.44.0` and `@types/eslint@8.44.0` are used in package-lock.json so that we can use `eslint@8.44.0` in our tests.

RunDevelopment (approved):

Issue #659, opened 2023-10-20
prefer-predefined-assertion does not recognize `(\.(?!$)|$))`


kurtextrem:
**Information:**
I used the playground to test (so I assume all latest?)

**Description**
Coming from this SO post: [stackoverflow.com/a/36760050](https://stackoverflow.com/a/36760050)
The regex

`^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)(\.(?!$)|$)){4}$`

got optimized (by hand) to:

`^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$`

so, `partbefore(\.(?!$)|$))` can be expressed as `(partbefore\.?\b)`, avoiding the negative lookahead, making the regex both smaller and faster: [esbench.com/bench/65326a147ff73700a4debb7d](https://esbench.com/bench/65326a147ff73700a4debb7d).

I assume `prefer-predefined-assertion` is meant to capture this, but it does not:
https://ota-meshi.github.io/eslint-plugin-regexp/playground/#eJwVydEKgjAUxvFXOR28cNBail60CB+hiy6dgbRjDHSOaVI5370FH/x/8K34GDWhRCHg6mYzmC9BazXczOB6032UVXZpPXiCC4C4p5WMy8v6yMsm/BlVNCFTSoc646eGVZEsPkod0mqXsJAwthZbIpYz7tG/eppQrujpSW8nnKeOPI/R1BlLmrfTRH42o0WZb9sP02szMA==

(by the way, if you turn on all rules, it errors for that regex: ![image](https://github.com/ota-meshi/eslint-plugin-regexp/assets/502146/97f1256a-111f-44a0-b484-2fa0319d259e))

ota-meshi:
Thank you for posting this issue!

However, I don't think `partbefore(\.(?!$)|$)` and `(partbefore\.?\b)` are equivalent.
The following script returns a different result.

```js
const s = 'partbefore.'
console.log(/partbefore(\.(?!$)|$)/.test(s), /(partbefore\.?\b)/.test(s)) // false true
```

I think there probably needs to be a `$` after the quantifier, but I haven't yet figured out how that can be optimized.

kurtextrem:
Yeah, looks like this is equivalent:
`console.log(/^partbefore(\.(?!$)|$)$/.test(s), /^(partbefore\.?\b)$/.test(s))`

[...]

PR #660, opened 2023-10-20
Fixed crash for v flag in `regexp/optimal-quantifier-concatenation`

ota-meshi:
related to #659

### DIFF
--- a/lib/rules/optimal-quantifier-concatenation.ts
+++ b/lib/rules/optimal-quantifier-concatenation.ts
@@ -68,9 +68,7 @@ function getSingleConsumedChar(
     element: Element | Alternative,
     flags: ReadonlyFlags,
 ): SingleConsumedChar {
-    const empty = flags.unicodeSets
-        ? EMPTY_UNICODE_SETS
-        : flags.unicode
+    const empty = flags.unicode || flags.unicodeSets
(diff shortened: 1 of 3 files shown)


lib/rules/optimal-quantifier-concatenation.ts
```diff
@@ -64,7 +68,11 @@ function getSingleConsumedChar(
     element: Element | Alternative,
     flags: ReadonlyFlags,
 ): SingleConsumedChar {
-    const empty = flags.unicode ? EMPTY_UNICODE : EMPTY_UTF16
+    const empty = flags.unicodeSets
+        ? EMPTY_UNICODE_SETS
+        : flags.unicode
+        ? EMPTY_UNICODE
+        : EMPTY_UTF16
```
RunDevelopment, line 75: The maximum character for Unicode and UnicodeSets mode is the same, so `EMPTY_UNICODE.char === EMPTY_UNICODE_SETS.char`. So we don't actually need `EMPTY_UNICODE_SETS` at all.

```suggestion
    const empty = flags.unicode || flags.unicodeSets
        ? EMPTY_UNICODE
        : EMPTY_UTF16
```

RunDevelopment (approved):

PR #663, opened 2023-10-20
Improve detection for `regexp/no-useless-assertion`

### LINKED ISSUE #661: Make `no-useless-assertion` detect `(?!$)$`
**Description**
While looking at #659, I noticed that `no-useless-assertion` did not detect that `(?!$)$` will always reject. We should fix that.

RunDevelopment:
Fixes #661.

The new condition for always accepting is relatively simple. It's just the previous check, but it accounts for edges properly and allows more assertions. 

The complex part of this PR is the `isSingleCharacterAssertion` function. Implementing this function was surprisingly complex. Since we can't modify the AST, we have to handle some pretty complex cases. But even with those complex cases handled, `isSingleCharacterAssertion` is still not perfect and will return false negatives at times. Fortunately, false negatives aren't a problem for us.

### DIFF
--- a/.changeset/odd-oranges-matter.md
+++ b/.changeset/odd-oranges-matter.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": minor
+---
+
+Improve detection of useless assertions for `regexp/no-useless-assertion`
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
Wonderful! Thank you.

PR #664, opened 2023-10-20
Fixed minor mistake in message of `no-misleading-unicode-character`

RunDevelopment:
In the error message, a grapheme with multiple code points was said to have "multiple char codes". A very minor mistake.


ota-meshi (approved):
Thank you!

PR #668, opened 2023-10-21
Improved error message and docs for `regexp/no-useless-non-capturing-group`

RunDevelopment:
Just something small I noticed.


ota-meshi (approved):
LGTM! Thank you!

PR #683, opened 2023-12-17
Fix typos in docs 

Logicer16:
I've noticed two minor errors in the documentation. This simply addresses them.


ota-meshi (approved):
Thank you!!

PR #687, opened 2023-12-29
docs: replaces 🏳️‍🌈 emoji with 🏴‍☠️ emoji

ota-meshi:
I received the following message from GitHub:

> The below code is promoting the idea that families from the LGBT community should not be seen well which we found  to be in violation of our Acceptable Use Policies. 
> 
> ```
> /* ✗ BAD */
> ❇️🏳️‍🌈👨‍👩‍👦]
> ```

I know we don't intend such a discriminatory meaning, but for some people to think it's discriminatory, that's not what I mean either, so I fixes the documentation.

For the record, I would like to strongly emphasize that this pull request to replace the flag is not intended to exclude LGBT people.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,8 +15,8 @@
     "css.validate": false,
     "typescript.tsdk": "./node_modules/typescript/lib",
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true,
-        "source.fixAll.stylelint": true
+        "source.fixAll.eslint": "explicit",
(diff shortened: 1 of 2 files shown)


RunDevelopment:
This seems like a false positive of some automated system, but I can also understand why even a human reviewer/moderator might have thought "better safe than sorry" and sent that message. Honestly, I find this a little funny as well, since the reasoning seems to be: the word "bad" in proximity to a rainbow = potentially discriminatory. A little simplistic, but I suspect that the rate of true positives is high enough, unfortunately.

PR #703, opened 2024-03-09
test: fix eslint v9 testcases

ota-meshi:


RunDevelopment:
Oh god, that must have been a lot of work. Thank you so much!

On that note: is there a tool to automatically update the expected output? Doing this by hand has always been annoying. If not, is there a snapshot-based testing solution for ESLint plugins?

RunDevelopment (approved):

ota-meshi:
>On that note: is there a tool to automatically update the expected output? Doing this by hand has always been annoying. If not, is there a snapshot-based testing solution for ESLint plugins?

Hmm. I created a mechanism to semi-automatically generate test results with eslint-plugin-svelte, but I don't know if it can be used successfully with eslint-plugin-regexp. It might be easier to create a separate mechanism for eslint-plugin-regexp 🤔 

https://github.com/sveltejs/eslint-plugin-svelte/blob/main/tests/fixtures/rules/no-extra-reactive-curlies/invalid/test01-errors.yaml
https://github.com/sveltejs/eslint-plugin-svelte/blob/main/CONTRIBUTING.md#test-the-rule

PR #708, opened 2024-03-11
use eslint-plugin-eslint-rule-tester

ota-meshi:
I'm not sure if this will work yet, but I think merging this PR will allow test case expectations to update automatically if you're using VSCode. Related to https://github.com/ota-meshi/eslint-plugin-regexp/pull/703#issuecomment-1987047177

See https://github.com/ota-meshi/eslint-plugin-eslint-rule-tester.

### DIFF
--- a/.eslintrc-editor-only.js
+++ b/.eslintrc-editor-only.js
@@ -1,0 +1,11 @@
+"use strict"
+
+module.exports = {
+    extends: [require.resolve('./.eslintrc.js')],
+    overrides: [
+        {
+            files: ["tests/lib/rules/*"],
+            extends: ['plugin:eslint-rule-tester/recommended-legacy']
+        },
+    ],
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,9 @@
         "jsonc",
         "yaml"
     ],
+    "eslint.options": {
+        "overrideConfigFile": "./.eslintrc-editor-only.js"
+    },
     "typescript.validate.enable": true,
     "javascript.validate.enable": false,
     "vetur.validation.script": false,
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
                 "eslint-doc-generator": "^1.7.0",
                 "eslint-plugin-eslint-comments": "^3.2.0",
                 "eslint-plugin-eslint-plugin": "^5.2.1",
+                "eslint-plugin-eslint-rule-tester": "^0.1.1",
                 "eslint-plugin-json-schema-validator": "^4.6.1",
                 "eslint-plugin-jsonc": "^2.0.0",
                 "eslint-plugin-n": "^16.0.0",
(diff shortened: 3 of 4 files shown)


RunDevelopment:
Sorry for the delay! 

Wow, that is an interesting idea. This integrates well into any project that uses ESLint's `RuleTester` and solves the problem of manually updating the expected values. The main drawback I see is that it's probably difficult for the plugin to actually run tests.

Also, I just checked out this branch and installed the plugin, but it didn't work for me when I played around in `tests\lib\rules\match-any.ts`.

---

I originally wanted to write this in our discussion in #703, but you were too fast for me :)

I never liked the format for expected values `RuleTester` uses, so I probably would have made a library to generate snapshots like this instead:

```
Options:
- report: interesting
  reportExponentialBacktracking: potential

Code:
  const foo3 = /(?:ac?|\wb?)a/.source;
                       ^~~~ 1
  const bar = RegExp(`^(?:${foo3})+$`);
                            ^~~~ 2

Output:
  const foo3 = /(?:ac?|\wb?)a/.source;
  const bar = RegExp(`^(?:${somethingElse})+$`);

Errors:
- message: Unexpected useless alternative. This alternative is a strict subset of 'ya?ml' and can be removed.
  suggestions:
  - message: Remove the 'yml' alternative.
    output:
      var re = /(?:ya?ml)/
- message: Some other error without suggestions. I don't know what else to write here.
```

Basically, it shows the location+range of errors with ASCII drawings and uses a YAML-like structure to show error details. The ASCII drawings means that we don't have to store line/column in the error details, and we get a visual representation of what the error will be like in IDEs.

As for actually writing tests, I would closely mirror the API of `RuleTester`. Only the `output` and `errors` props of invalid test cases would have to be removed. Like this:

<details>
<summary>Modified version of no-useless-backreference</summary>

```ts
import { SnapshotRuleTester } from "eslint-snapshot-rule-tester"
import rule from "../../../lib/rules/no-useless-backreference"

const tester = new SnapshotRuleTester({
    languageOptions: {
        ecmaVersion: "latest",
        sourceType: "module",
    },
})

tester.run("no-useless-backreference", rule as any, {
    valid: [
        "/.(?=(b))\\1/",
        "/(?:(a)|b)\\1/",
        "/(a)?\\1/",
        "/(a)\\1/",
        "/(?=(a))\\w\\1/",
        "/(?!(a)\\1)/",
    ],
    invalid: [
        "/(b)(\\2a)/",
        "/(a\\1)/",

        "/(\\b)a\\1/",
        "/([\\q{}])a\\1/v",
        "/(\\b|a{0})a\\1/",

        "/(a)b|\\1/",
        "/(?:(a)b|\\1)/",
        "/(?<=(a)b|\\1)/",

        "/\\1(a)/",
        "/(?:\\1(a))+/",

        "/(?<=(a)\\1)b/",

        "/(?!(a))\\w\\1/",
        {
            code: "/(?!(?!(a)))\\w\\1/",
            options: [{ foo: 3 }]
        },
    ],
})
```

</details>

The main advantages of this snapshot approach are that the snapshot syntax isn't limited by JS strings and that it's easier to add invalid test cases (use a source code string, just like for valid test cases). The main downsides of this approach are that another dependency is needed to run tests and that existing invalid test cases need to be modified.

Well, your approach is most likely more practical, but I wanted to share this anyway because I like snapshots :)

ota-meshi:
However, I'm not sure why it doesn't work with `tests\lib\rules\match-any.ts`. It works fine on my local.


https://github.com/ota-meshi/eslint-plugin-regexp/assets/16508807/c115cd37-9965-4c4f-a015-5ad4a2ef84c0

RunDevelopment:
> Are you already working on `eslint-snapshot-rule-tester`?

I am now :) I'll make a PR tomorrow.

PR #709, opened 2024-03-18
Use snapshots in rule tests

RunDevelopment:
This took waaaaay too long to make because I couldn't decide on a snapshot format. That said, I'm quite happy with the result.

This PR uses [`eslint-snapshot-rule-tester`](https://github.com/RunDevelopment/eslint-snapshot-rule-tester) to get us snapshot testing for invalid test cases. This means (1) that we don't have to manually update expected values when a rule changes, and (2) that we see the full range information of errors. I changed all rule tests to use snapshot tests, so a lot of files changed. Basically, I just used some find & replace with regexes to remove the `errors` and `output` fields.


ota-meshi (approved):
Wow! Snapshot files are so easy to view and amazing! Thank you so much!

PR #710, opened 2024-03-20
Refactor utils

### LINKED ISSUE #576: Module organization in `lib/utils`?
So I have a question about code organization in `lib/utils`: how does it work/what is the structure behind it?

I wondered this for a long time. From my perspective, it just looks like a mess. Of course, I never understood the file layout, so I very much contributed to that mess.

What confuses me the most are `index.ts` files with function/class/type definitions. `index.ts` files in my projects typically only re-export other modules to define an API. I pretty much never put definitions in `index.ts` files, because e.g. `a/b.ts` cannot import a function from `a/index.ts` (cyclic import).

---

As an example: I want to add new util function:
```ts
function assertNever(value: never): never {
  throw new Error("Unsupport value: " + value)
}
```

This might be most util of util method. It has no dependencies, and I will need it in a lot of places (see #575 for more info).

Importantly, many util methods also need this, so I can't define it in `lib/util/index.ts` (cyclic import). But I also don't really want to make a `lib/util/util.ts`. Where should I put this function?

Also, we have around 5 copies of `assertNever` in various files already.

---

Maybe a bit of background. In my JS/TS projects, I typically go for a flat hierarchy for internal modules (= files that don't define functions/classes/types of the public API). So each of those internal modules just exports their functions/classes/types and that's it. Those internal modules might be in different folders, but there is no hierarchy between them. So internal modules never re-export things from other internal modules and there are no internal `index.ts` files.

I do this because it makes it easy to extract out common functionality into a module. I don't have to worry about internal hierarchies and can just create a new file.

Again, I'm only talking about internal modules. I use `index.ts` files and re-exports to define the public API of a package like everybody else. Although my `index.ts` files only re-export other modules, I don't define any functions/classes/types in them.

RunDevelopment:
Resolves #576

Unlike what I said in the issue, I decided against the flat layout I usually go for. I did this for 2 reasons:

1. Minimizing the diff. I already touched enough files as is, so I didn't want to change every rule we have.
2. Managing complexity. We have a lot of code in `lib/utils`, so using re-exports via `index.ts` can help with managing its complexity by defining internal boundaries (everything *not* re-exported is private). The best example for this is `type-tracker`. This directory contains a lot of code, but the API of the module is minimal (only 1 function and 1 type).

With that out of the way, here's what I changed:

- Renamed `lib/utils/rules.ts` to `lib/all-rules.ts`. I did not like that something inside utils imported every rule we have, so I moved it outside.
- I moved a lot of code from `lib/utils/index.ts` into different files. `lib/utils/index.ts` did a little bit of everything, so I "tore" it apart.
- Added `lib/utils/regex-syntax.ts` for functions and constants dealing with regex source strings directly. This contains things like `RESERVED_DOUBLE_PUNCTUATOR_CHARS` and `isOctalEscape`.
- I interpreted `lib/utils/regexp-ast/` to contain the util functionality for analyzing regexpp ASTs and moved some functions into it. Things like `quantToString` and `hasCapturingGroup`.
- I added `lib/utils/refa.ts` for refa-related code.
- I also tried to deduplicate some code by using existing utility functions.

Since this is a pure refactoring PR, it contains no functional changes.

---

Also, I noticed that don't have an ESLint rule for ordering `import`s. Should I add one after this PR?


ota-meshi (approved):
LGTM! Thank you for all your work!

> Also, I noticed that don't have an ESLint rule for ordering imports. Should I add one after this PR?

Yeah. We can introduce [eslint-plugin-import] to this repo 👍 

[eslint-plugin-import]: https://github.com/import-js/eslint-plugin-import

PR #722, opened 2024-04-06
Add `regexp/unicode-property` rule

### LINKED ISSUE #720: Prefer long unicode class names over the short ones
**Motivation**

The long hand names are more readable.

**Description**

Same as above.

**Examples**

```js
/* ✓ GOOD */
/\p{Letter}/v

/* ✗ BAD */
/\p{L}/v
```

```js
/* ✓ GOOD */
 /\p{Script=Greek}/v

/* ✗ BAD */
/\p{sc=Greek}/v
```

RunDevelopment:
Fixes #720

This PR adds a new rule that allows users to enforce the naming of Unicode properties. It has 3 main features:

1. Removing/adding `gc=`/`General_Category=` keys, e.g. `\p{gc=L}` -> `\p{L}`. These prefixes are unnecessary, because the values of the `General_Category` property can be accessed without the key.
2. Enforcing long or short keys for `General_Category`/`gc`, `Script`/`sc`, and `Script_Extensions`/`scx`.
3. Enforcing long or short names of values and binary properties. E.g. `\p{L}` -> `\p{Letter}` and `\p{Hex}` -> `\p{Hex_Digit}`.

All of these feature can be individually configured and turned off by the user. The `regexp/unicode-property` is **not** included in our `recommended` config, because this rule only enforces a specific style.

### Default configuration

The default configuration is the following:
```json
{
    "generalCategory": "never",
    "key": "ignore",
    "property": {
        "binary": "ignore",
        "generalCategory": "ignore",
        "script": "long",
    }
}
```

This means that, by default, the rule will (1) remove `General_Category`/`gc` keys (e.g. `\p{gc=L}` -> `\p{L}`) and (2) enforce long names for values of the `Script` and `Script_Extensions` properties (e.g. `\p{sc=Kana}` -> `\p{sc=Katakana}`).

I chose a minimal configuration because I didn't want to make the rule generate a lot of error for people trying to adapt the rule. I think the 2 effects work well in any code base, no matter what style they usually prefer. (1) simply removes an unnecessary prefix to "simplify" the regex, and (2) prevents the use of the (IMO) horrible aliases for scripts.

### Unicode data

Since I needed the data for the mapping between aliases to implement this rule, I had to make the choice between taking a dependency (e.g. `@unicode/unicode-15.0.0`) or including the relevant data in the source files of this project.

I chose *against* adding a dependency, because it was easy enough to get the data I needed and because most of ``@unicode/unicode-15.0.0`` would be dead weight to us.

However, the data I included is used through an API (the `AliasMap` class), so we can easily switch to using a dependency without needing to change the `regexp/unicode-property` rule.


docs/rules/unicode-property.md
```diff
@@ -0,0 +1,241 @@
...
+  /* eslint regexp/unicode-property: ["error", { generalCategory: "ignore" }] */
+
+  var re = /\p{Letter}/u;
+  var re = /\p{gc=Letter}/u;
+  var re = /\p{General_Category=Letter}/u;
+  ```
+
+  </eslint-code-block>
+
+### `key: "short" | "long" | "ignore"`
```
fisker, line 103: Should we support "medium"? https://unicode-org.github.io/icu/userguide/strings/unicodeset.html#property-values

docs/rules/unicode-property.md
```diff
@@ -0,0 +1,241 @@
...
+  var re = /\p{General_Category=Letter}/u;
+  var re = /\p{sc=Greek}/u;
+  var re = /\p{Script=Greek}/u;
+  var re = /\p{scx=Greek}/u;
+  var re = /\p{Script_Extensions=Greek}/u;
+  ```
+
+  </eslint-code-block>
+
+### `property: "short" | "long" | "ignore" | object`
```
fisker, line 157: We may want include a note to explain that "long" doesn't mean in length since "Han"(long) sometimes shorter than "Hani"(short)

docs/rules/unicode-property.md
RunDevelopment: We already do with `generalCategory: "always", key: "short"`. This will enforce `\p{gc=X}`. So it's already possible to configure this rule to act like the "medium" from the page you linked.

I also don't think that a "medium" option really makes sense, because it only applies to the `General_Category` property as it is the only property (in JS) that can have 3 forms (`\p{X}`, `\p{gc=X}`, `\p{General_Category=X}`).

docs/rules/unicode-property.md
RunDevelopment: Thank you for pointing this out. I documented exceptions to the notation of short and long names in a note.

ota-meshi (commented):
Looks great to me! Thank you!

PR #724, opened 2024-04-07
Publish type declarations

### LINKED ISSUE #723: Publish type declarations
**Description**

I type-check my `eslint.config.js` file (and I'm sure many people do as well) and it would be great if I don't get these errors in them:
```log
Could not find a declaration file for module 'eslint-plugin-regexp'. './node_modules/eslint-plugin-regexp/dist/index.js' implicitly has an 'any' type.
```
This plugin is already written in TypeScript so it shouldn't be too much trouble to add the type declarations to the published package.

RunDevelopment:
Fixes #723

This PR makes our build process emit type declarations (`.d.ts` files). Since we publish the entire `dist/` folder, they will be included when publishing without any further config changes. In my tests, types declarations were correctly imported, so this seems to work.

One thing to note is that this PR publishes a lot more `.d.ts` files than necessary. All files in `lib/rules` and `lib/utils` will have their type declarations published as well. I don't think there are any real downsides that come from this, but I think it's something we should be aware of.

### DIFF
--- a/.changeset/mean-peas-give.md
+++ b/.changeset/mean-peas-give.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": patch
+---
+
+Publish type declarations
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
         "node": "^18 || >=20"
     },
     "main": "dist/index.js",
+    "types": "dist/index.d.ts",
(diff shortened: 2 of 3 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #735, opened 2024-04-19
chore: remove intl-segmenter-polyfill

ota-meshi:
This PR removes intl-segmenter-polyfill.
intl-segmenter-polyfill was used on our site. This is because Intl.Segmenter did not work in FireFox.
However, FireFox now supports Intl.Segmenter, so intl-segmenter-polyfill is no longer needed.

https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/125

### DIFF
--- a/docs/.vitepress/theme/index.ts
+++ b/docs/.vitepress/theme/index.ts
@@ -19,10 +19,6 @@ const theme: Theme = {
     async enhanceApp(ctx) {
         DefaultTheme.enhanceApp(ctx)
 
-        if (typeof Intl.Segmenter === "undefined") {
-            await setupIntlSegmenter()
-        }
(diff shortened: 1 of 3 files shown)


RunDevelopment:
This update has been released 3 days ago. Should we wait a little to give people time to update their browser?

ota-meshi:
I don't need it right away, so it can wait :+1:  Should we wait about a month?

RunDevelopment:
A month sounds good. The release cycle of Firefox seems to be 4 weeks, so giving people one version as a buffer should be good.

ota-meshi:
I forgot about this PR, but it's been a few months so I'll merge it.

PR #740, opened 2024-05-19
Generate string literal types for configs

karlhorky:
Hi there, first of all thanks for this ESLint plugin, great to learn about regular expression antipatterns!

Using the `tseslint.config()` config helper function from [`typescript-eslint`](https://typescript-eslint.io/) leads to an error:

```ts
import tseslint from 'typescript-eslint';

const config = tseslint.config(
  regexp.configs['flat/recommended'] // 💥 Argument of type 'typeof import("/Users/k/p/project/node_modules/eslint-plugin-regexp/dist/configs/flat/recommended")' is not assignable to parameter of type 'ConfigWithExtends'.
                                     // (see more below)
)
```

Error message:

```
Argument of type 'typeof import("/Users/k/p/project/node_modules/eslint-plugin-regexp/dist/configs/flat/recommended")' is not assignable to parameter of type 'ConfigWithExtends'.
  Types of property 'rules' are incompatible.
    Type '{ "no-control-regex": string; "no-misleading-character-class": string; "no-regex-spaces": string; "prefer-regex-literals": string; "no-invalid-regexp": string; "no-useless-backreference": string; ... 60 more ...; "regexp/use-ignore-case": string; }' is not assignable to type 'Partial<Record<string, RuleEntry>>'.
      Property '"no-control-regex"' is incompatible with index signature.
        Type 'string' is not assignable to type 'RuleEntry | undefined'.ts(2345)
```

This is because [the types for `typescript-eslint` require a string of `'error'`,  `'off'` or `'warn'`](https://github.com/typescript-eslint/typescript-eslint/blob/8d92ba8533e65360877de1af12979c42b6c836e2/packages/utils/src/ts-eslint/Config.ts#L11)

To avoid the type mismatch when using `typescript-eslint`, this PR aims to generate string literal config property values instead of [the current property values of `string`](https://unpkg.com/browse/eslint-plugin-regexp@2.5.0/dist/configs/rules/recommended.d.ts)

### DIFF
--- a/.changeset/grumpy-waves-warn.md
+++ b/.changeset/grumpy-waves-warn.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": minor
+---
+
+Improved compatibility of type information with typescript-eslint in config.
--- a/lib/configs/rules/all.ts
+++ b/lib/configs/rules/all.ts
@@ -1,7 +1,8 @@
 import { rules as ruleLint } from "../../all-rules"
+import type { SeverityString } from "../../types"
 import { rules as recommendedRules } from "./recommended"
 
-const all: Record<string, string> = {}
+const all: Record<string, SeverityString> = {}
 for (const rule of ruleLint) {
     all[rule.meta.docs.ruleId] = "error"
 }
--- a/lib/configs/rules/recommended.ts
+++ b/lib/configs/rules/recommended.ts
@@ -1,4 +1,6 @@
-export const rules = {
+import type { SeverityString } from "../../types"
+
+export const rules: Record<string, SeverityString> = {
     // ESLint core rules
     "no-control-regex": "error",
     "no-misleading-character-class": "error",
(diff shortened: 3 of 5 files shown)


ota-meshi (changes requested):
Thank you for this PR!
That file is automatically updated by `tools/update-rulesets.ts`, so could you please fix that as well?
Also, if there is a similar problem in `lib/configs/rules/all.ts`, could you please fix it as well?

karlhorky:
@ota-meshi thanks for the feedback! I've added a `SeverityString` type and updated the files you mentioned in 7efd9e3d6ebc93043d0b8bb4a9114063e316e317

ota-meshi (approved):
LGTM! Thank you!

karlhorky:
Thanks for the review, merge and publish of [`eslint-plugin-regexp@2.6.0`](https://github.com/ota-meshi/eslint-plugin-regexp/releases/tag/v2.6.0) 🙌 

I can confirm that this fixed the type error mentioned above with `typescript-eslint` types - no type error anymore 👍 :

Screenshot 2024-05-21 at 16 55 58

PR #752, opened 2024-06-28
Add support for ES2025 duplicate named capturing groups

ota-meshi:
This PR updates regexpp and some rules to add support for ES2025 duplicate named capturing groups.

PR #754, opened 2024-07-16
chore: fix workflows error

ota-meshi:


ota-meshi:
The test workflow error is fix in #752.
I will merge this to fix the cron workflow error.

PR #774, opened 2024-11-03
docs: simplify import in flat config

rakleed:


ota-meshi:
It either works, but I prefer the current documentation because I think I'll make it named exports in the future when I make this plugin an ESM.

PR #776, opened 2024-11-14
Correct capturing group not found grammar

adamlui:


ota-meshi (approved):
Thanks for the PR!
(CI is failing but that has nothing to do with this PR.)

PR #806, opened 2025-02-24
ci: add node 22 to test matrix

Fdawgs:
As this is going to be the next LTS it makes sense to test against it. :)


ota-meshi (approved):
thank you!

PR #808, opened 2025-03-01
fix: `no-unused-capturing-group` rule match error

btea:
When the regular expression of the following group is used in the match, the previous group should also be considered to be used.

### DIFF
--- a/docs/rules/no-unused-capturing-group.md
+++ b/docs/rules/no-unused-capturing-group.md
@@ -26,6 +26,7 @@ This rule reports unused capturing groups.
 
 /* ✓ GOOD */
 var replaced = '2000-12-31'.replace(/(\d{4})-(\d{2})-(\d{2})/, '$1/$2/$3') // "2000/12/31"
+var replaced = '2000-12-31'.replace(/(\d{4})-(\d{2})-(\d{2})/, '$3') // "31"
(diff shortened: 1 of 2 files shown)


ota-meshi:
I think the rule is working correctly now.
Could you open an issue and explain in more detail what you want to do?

btea:
For example, in the example I added, I only used group 3 when replacing, and the current rules will prompt that group 1 and group 2 are not used. But group 3 is based on group 1 and group 2, so I think there should be no error prompt.

```js
var replaced = '2000-12-31'.replace(/(\d{4})-(\d{2})-(\d{2})/, '$3') // "31"
```

ota-meshi:
In fact the first and second capturing groups in your regex are unused.

btea:
But they are necessary and cannot be changed to non-capturing groups, otherwise the third group will not be established.

ota-meshi:
I think you need to do this:

```js
var replaced = '2000-12-31'.replace(/(?:\d{4})-(?:\d{2})-(\d{2})/, '$1')
```

Is there a reason why this isn't possible?
And could you open an issue?

btea:
Oh, thanks for the example. I seem to have fallen into a thinking error. 👍 

In this case, can we make the error message more detailed?

PR #830, opened 2025-05-11
style: enable `unicorn/prefer-string-raw`

fisker:


eslint.config.mjs
```diff
@@ -242,4 +246,11 @@ export default [
             "n/no-missing-import": "off",
         },
     },
+    {
+        ignores: ["lib/**/*", "**/*.json"],
```
fisker, line 250: Didn't run on `lib/`, I'm not it's wanted.

ota-meshi (approved):
Thank you!!!

PR #992, opened 2026-06-24
Handle matchAll().toArray() capture usage

### LINKED ISSUE #986: [no-unused-capturing-group] Doesn't handle `toArray`
**Information:**
- ESLint version: 10.5.0
- `eslint-plugin-regexp` version: 3.1.0

`no-unused-capturing-group` correctly handles `[...query.matchAll(/as(d)/)]` (i.e. it *doesn't* report an error) but it does not handle (i.e. it gives a false-positive warning) `query.matchAll(/as(d)/).toArray()` which is exactly the same thing.

morgan-coded:
`no-unused-capturing-group` could report captures as unused when match results were consumed through `matchAll(...).toArray()`.
This treats iterator-helper `toArray()` as consuming the match arrays, matching the existing behavior for other result-consumption paths. I checked the red-to-green repro, the focused rule suite with 68 passing, typecheck, build, lint, and `git diff --check`.
Fixes #986

### DIFF
--- a/tests/lib/rules/no-unused-capturing-group.ts
+++ b/tests/lib/rules/no-unused-capturing-group.ts
@@ -133,6 +133,9 @@ tester.run("no-unused-capturing-group", rule as any, {
         const matches = [...('2000-12-31 2000-12-31'.matchAll(/(\d{4})-(\d{2})-(\d{2})/g))]
         `,
         String.raw`
+        console.log('asd'.matchAll(/as(d)/gu).toArray())
+        `,
+        String.raw`
         const bs = [...'abc_abc'.matchAll(/a(b)/g)].map(m => m[1])
         `,
         String.raw`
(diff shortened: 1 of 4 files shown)


tests/lib/rules/no-unused-capturing-group.ts
```diff
@@ -133,6 +133,9 @@ tester.run("no-unused-capturing-group", rule as any, {
         const matches = [...('2000-12-31 2000-12-31'.matchAll(/(\d{4})-(\d{2})-(\d{2})/g))]
         `,
         String.raw`
+        console.log('asd'.matchAll(/as(d)/gu).toArray())
+        `,
+        String.raw`
```
ota-meshi, line 138: Could you please add the following to the valid test cases,

```js
        String.raw`
        const bs = 'abc_abc'.matchAll(/a(b)/g).toArray().map(m => m[1])
        `,
```

and add the following to the invalid test cases?

```js
        {
            code: String.raw`
            const bs = 'abc_abc'.matchAll(/a(b)/g).toArray().map(m => m[0])
            `,
            options: [{ fixable: true }],
        },
```

ota-meshi (changes requested):
Thank you for this PR!

ota-meshi (approved):
LGTM! Thank you!

PR #1009, opened 2026-07-24
docs: explain escape-case compatibility

### LINKED ISSUE #743: Rule `letter-case` does not fix `unicodeEscape`
**Information:**
- ESLint version: 8.57.0
- `eslint-plugin-regexp` version: 2.6.0

**Description**

With this rule elabled:

```js
"regexp/letter-case" : ["error", {
  caseInsensitive : "ignore",
  unicodeEscape : "lowercase",
  hexadecimalEscape : "lowercase",
  controlEscape : "uppercase"
}]
```

The following code:

```js
const REG2 = /\u036F/u;
```

is reported as error:

> error  '\u036F' is not in lowercase               regexp/letter-case

but should be automatically fixed to:

```diff
- const REG2 = /\u036F/u;
+ const REG2 = /\u036f/u;
```

Note that [the rule doc](https://ota-meshi.github.io/eslint-plugin-regexp/rules/letter-case.html) specifies that:

> 🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).

ColumbusLabs:
## Summary

Documents the default conflict between `regexp/letter-case` and `unicorn/escape-case`.

The new compatibility note shows how to align both Unicode and hexadecimal escape handling with Unicorn's default uppercase style. Configuring both properties is necessary to avoid circular autofixes for `\u` and `\x` escapes.

Closes #743.

## Validation

- `npm run build`
- `npm run update`
- `npm run lint:docs`
- `npm run docs:build`
- Verified the documented configuration against Unicode, hexadecimal, and control escapes
- `git diff --check`

### DIFF
--- a/docs/rules/letter-case.md
+++ b/docs/rules/letter-case.md
@@ -60,6 +60,23 @@ var foo = /\ca/
   - `hexadecimalEscape` ... Specifies the letter case when the hexadecimal escapes. Default is `"lowercase"`.
   - `controlEscape` ... Specifies the letter case when the control escapes (e.g. `\cX`). Default is `"uppercase"`.
 
+### Compatibility with `unicorn/escape-case`
+
+The default `"lowercase"` values for `unicodeEscape` and `hexadecimalEscape`
+conflict with the default `"uppercase"` style of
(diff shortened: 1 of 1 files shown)


ota-meshi (approved):
Thank you!

PR #1012, opened 2026-07-26
chore: update Prettier to v3.9.6

fisker:


lib/utils/type-tracker/type-data/function.ts
```diff
@@ -187,12 +187,10 @@ const getPrototypes: () => {
...
-    createObject<
-        {
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type -- ignore
-            [key in keyof Function]: TypeInfo | null
-        }
-    >({
+    createObject<{
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type -- ignore
+        [key in keyof Function]: TypeInfo | null
+    }>({
```
fisker, line 193: Regression in Prettier 3.9.0 https://github.com/ota-meshi/eslint-plugin-regexp/commit/407214d528352bce56916b963e00b09822541e6d#diff-724e8f64653814cfa42c23478286bd18fd939e08c5742179b2b531e80357ed4bL190-L192 , also the reason I made this change to Prettier https://github.com/prettier/prettier/blob/main/CHANGELOG.md#typescript-ignore-comments-inside-mapped-type-when-checking-type-parameter-comments-19572-by-fisker

ota-meshi (approved):
Thank you!

PR #1042, opened 2026-09-03
Handle matchAll().take() and .drop() capture usage

NotAFlightRisk:
`no-unused-capturing-group` reports group 1 on this, and the suggestion it offers rewrites the regex to `/as(?:d)/gu`, which breaks the `m[1]` sat under it:

```js
for (const m of 'asd'.matchAll(/as(d)/gu).take(2)) console.log(m[1])
```

The same line with `.toArray()` instead is quiet, so its just the name. `take` and `drop` are the other two iterator helpers with no `Array.prototype` twin, which is why they cant arrive through `keyof unknown[]` like most of that table does, and #992 added `toArray` on its own. I've put a negative control in the tests as well, because `.take(2).map(m => m[0])` should still report, and it does.

Small change, so feel free to close it if you'd rather leave the table as it is.

---

One I've left alone. Any unrecognised method on a matchAll result reads as unused, not just these two, so `'asd'.matchAll(/as(d)/gu).foo()` reports exactly the same. That's a behaviour change rather than a missing name though, and it seemed like your call and not mine.

### DIFF
--- a/.changeset/matchall-take-drop-captures.md
+++ b/.changeset/matchall-take-drop-captures.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-regexp": patch
+---
+
+Handle matchAll().take()/.drop() capture usage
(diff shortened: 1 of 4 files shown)


ota-meshi (approved):
LGTM! Thank you!

